PR #21, opened 2021-11-23
Add core & proxy REST endpoints

edd:
The current `main` only contains the data-node swagger.json. This PR adds the other 3 REST services we have

- Add core state swagger
- Add core swagger
- Add data-node proxy swagger

Note: This doens't automatically add them to the sidebar. That's why this is WIP.


edd:
@candida-d handing this over to you. I've:
- Added the other REST services to the sidebar
- Added a REST overview 
- Updated `.gitignore` so that the *generated* graphQL pages aren't committed. As they're built during deploy, it just leads to conflicts. This does mean that if you navigate in to the GraphQL section without running `./scripts/build.sh`, it'll look a bit empty. But if you run that script, or ignore the emptiness, it will be fine on deploy

What still needs doing:
- The 4 REST endpoint names were made up by me on the fly, and are not useful
- The rest/overview document is a *start*, but contains factual errors and the same confusing names as above.

Your mission, whether or not you choose to accept it, is to find good names for those endpoints that represent what they do, and fix the REST overview doc.

What is still left to do is make the sidebars in the API folder generally more useful. Between GQL/REST/Protodocs, all produced by different plugins, there's a slightly jarring difference between all of them. It's releasable with the menus as they are, but not particularly nice.

candida-d (approved):

PR #31, opened 2022-01-04
Make CLI lowercase in file name for easier redirects

candida-d:


edd (approved):
Nice! It's worth checking other folder names for caps and making them lower case, but that's a follow up. This appears to fix the issue 👍

PR #34, opened 2022-01-05
Update how rewards are divided up

candida-d:
Rewards are proportional to the amount of stake, and this PR makes that clear in the concepts.


edd (approved):
Reads well, and matches how I interpret [the relevant spec](https://github.com/vegaprotocol/specs-internal/blob/master/protocol/0061-simple-POS-rewards-SweetWater.md#calculation) 👍

PR #36, opened 2022-01-06
Add @easyops-cn/docusaurus-search-local

edd:
@candida-d does this solve your search needs? If it does it will need some minor CSS tweaks. I think it's good!

- Add [docusaurus-search-local](https://github.com/easyops-cn/docusaurus-search-local) dependency


candida-d:
Looks cool! Only trouble I've had with it is that when you get API documentation results, it seems to only be GraphQL(?) and because of how the autogen pages are rendered it's not clear that it is GraphQL you're seeing.

edd:
@candida-d This is good enough to go with, while still being imperfect. 

# What's not great
- API documentation consistency is poor. This is because we use different plugins for our REST, GRPC and GraphQL documentation:
  - REST is not indexed by search, because it doesn't really generate pages. When https://github.com/vegaprotocol/protos/issues/229 is done, this may become an option
  - GRPC is not indexed because it currently generated to a `/protodocs` folder. I suspect this can be addressed, but I haven't as part of this PR
  - GraphQL *is* indexed, but the search results for some things don't make it clear enough that the search result is for GraphQL. I've fixed this slightly by updating the title on the schema root page to be 'GraphQL schema'. so that when you land on it you realise you're looking at GraphQL docs. However if you search for `Reward`, the search result would be clearer if it was prefixed with 'GraphQL', but the only way I can see to do that has downsides too. 
- The last point there would be a showstopper if it wasn't for the other two points. Also these downsides only apply to API docs - the results are great for everything else.

# What is great
- Everything else

edd:
I've unhidden the title for protobuff docs, which makes them slightly more distinguishable from the GraphQL docs. I think there's scope for improvement, but I also think this is good enough to merge, wait until the aforementioned improvement for REST and revisit.

candida-d:
Thanks @sam-keen! So much easier to read now.

edd:
Nice work @sam-keen! @candida-d do we need to address other points? Or to put it another way: what happens next?

PR #40, opened 2022-01-13
Isolate key guide + tidying up

candida-d:


docs/tools/vega-wallet/cli-wallet/latest/guides/isolate-keys.md
```diff
@@ -0,0 +1,250 @@
...
+---
+
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
+If you want to move a key pair from a standard wallet into a wallet that can only sign and verify transactions, use the `key isolate` command. Isolating a key pair will extract a chosen key pair from one wallet and create an isolated wallet with only that key pair.
+
+This optional step creates an extra layer of security, particularly for validators and others who are concerned about their wallets being compromised. 
+
+An isolated wallet can only contain a single key pair that has been stripped from its cryptographic node. A wallet without a cryptographic node, which is used to create keys, minimises the impact of that wallet being compromised, as it makes it impossible to retrieve or generate keys.
```
ValentinTrinque, line 14:
```suggestion
An isolated wallet can only contain a single key pair that has been stripped from its cryptographic node. A wallet without a cryptographic node minimises the damages if the wallet has been compromised, as it makes it impossible to generate keys or retrieve any other keys than isolated one. 
```

It's very badly worded, but I think sentence is too heavy and convoluted. We should find a better way to explain why not having this node is beneficial to security.

ValentinTrinque (changes requested):

docs/tools/vega-wallet/cli-wallet/latest/guides/isolate-keys.md
```diff
@@ -0,0 +1,250 @@
...
+      "vega_pub_key": "ISOLATED_PUBLIC_KEY",
+      "ethereum_address": "MY_ETHEREUM_ADDRESS",
+      "tm_pub_key": "TENDERMINT_PUBLIC_KEY",
+      "info_url": "VALIDATOR_URL",
+      "country": "XX",
+      "name": "NODE_NAME",
+      "avatar_url": ""
+    }
+  }
+```
```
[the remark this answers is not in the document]
candida-d: Thought you might say that. I put them together as a temporary option, to get it out without having to create a whole new section with just that one guide, as it's a little bit out of context without more guides. I'm willing to break them up if you think it's too confusing or irrelevant. Though... it will also take a little longer to get this one out if so :)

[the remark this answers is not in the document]
ValentinTrinque: Alright, then it's fine.

docs/tools/vega-wallet/cli-wallet/latest/guides/isolate-keys.md
candida-d: An isolated wallet can only contain a single key pair that has been stripped from its cryptographic node. If a wallet without a cryptographic node is compromised, the damage is minimised as it is impossible to generate keys or retrieve any keys other than the isolated one. 


How about that ^ @ValentinTrinque?

ValentinTrinque: It's honey to my ears.

ValentinTrinque (approved):

PR #43, opened 2022-01-20
Add validator score details to concepts

candida-d:
This PR covers the concept of validator scores, in the context of why rewards might not be distributed based on validators not being appropriately staked. 

The info should be correct for version 0.47.5.

### DIFF
--- a/docs/concepts/vega-chain.md
+++ b/docs/concepts/vega-chain.md
@@ -13,19 +13,8 @@ Vega runs on a delegated proof of stake blockchain. Participants -- validators a
 
 Everyone participating in keeping the network secure, robust and reliable, including nominators, is rewarded for keeping the network running. Not meeting the requirements of running the network can lead to penalties. 
 
-### Penalties 
-Validators that don't meet the requirements or prove to be bad actors will have rewards withheld. Nominators of a validator that doesn't meet the requirements will also receive fewer (or no) rewards. A validator's performance is calculated based on their **validator score**, and in the future their **performance score**. 
-
-### Validator score
-The validator score is calculated based on how much stake a validator has. If, at the end of an epoch, a validator does not have sufficient self-stake or has overall too much stake, then their validator score will be lowered. 
-
-<!--
-### ***Further reading***
-Link to spec for staking and validator rewards when publicly available. (valpol)
--->
-
 ### VEGA token
-Vega uses the VEGA ERC20 token for governance, which includes nominating validators, and creating and voting on governance proposals.
+Vega uses the VEGA ERC20 token for governance, which includes nominating validators to run nodes, and creating and voting on governance proposals.
 
 A VEGA token (or fraction) can be either unassociated or associated with a Vega key:
 
@@ -37,85 +26,6 @@ A VEGA token (or fraction) can be either unassociated or associated with a Vega 
 :::info
 A user's VEGA tokens must first be associated with a Vega key before they can be used for governance and staking.
 :::
-
-### Staking on Vega 
-Vega networks use the ERC20 token VEGA for staking. Staking requires the combined action of associating VEGA tokens (or fractions of a token) to the Vega staking bridge contract; and using those token(s) to nominate one or more validators. 
-
-**Epochs** 
-An epoch is a time period during which staking changes can be announced and then implemented. Changes that are announced in one epoch will only be executed in the following epoch (excepting 'un-nominate now' - see below). The length of an epoch is set by the network parameter `validators.epoch.length`. 
-
-### Nominating validators
-Using tokens to nominate validators keeps the decentralised network functioning. 
-
-Tokenholders can nominate validators to encourage a diverse set of reliable nodes running the network, and to give the community the opportunity to disincentivise and/or remove bad validators.
-
-When a tokenholder chooses a validator (or validators) to nominate with their tokens, the amount is immediately subtracted from their available balance, and is used at the start of the next epoch to actively nominate those validator(s). 
-
-:::info
-VEGA tokenholders can use [token.vega.xyz](https://token.vega.xyz) to associate their tokens and nominate validators. A Vega Wallet and Ethereum wallet are both required. CoinList custodial users should confirm with CoinList how staking works for them.
-:::
-
-### Maximum stake per validator
-Each validator has a maximum amount of stake that they can accept. During restricted mainnet, this will be the same amount for all validators. 
-
-The max stake per validator equation includes all nominations and un-nominations requested in the next epoch, and the current epoch's active nominations, divided by the optimal number of validators. The optimal number of validators is defined using network parameters. 
-
-It is calculated as: 
-
-```
-max stake per validator = (
-current total stake across all validators 
-   - total requested un-nominated stake 
-   + total requested nominated stake) 
-/ 
-(max(
-    reward.staking.delegation.minValidators, 
-    (number of validators 
-        / reward.staking.delegation.competitionLevel
-    )
-)
-```
-
-### Automatic nomination
-Automatic nomination is triggered when an individual tokenholder has manually nominated 95%+ of their associated tokens. At that point, any newly associated tokens will automatically be nominated to the same validators, in the same proportion.
-
-Exceptions to automatic nomination: 
-* If, ahead of the next epoch a participant uses their available tokens to nominate validators manually, that takes precedence over automatic nomination. 
-* For the epoch after un-nominating validators (see below), tokens are not auto-nominated, to provide time to change the delegation / remove tokens.  
-
-### Un-nominating validators
-Participants can remove their nomination at the end of an epoch, or immediately. The un-nominated tokens will be restored back to the participant's associated token balance.
-
-**Un-nominate towards the end of the epoch** 
-
-A participant can un-nominate towards the end of the current epoch, which means the stake is not used for the validator from the following epoch. 
-
-The action is announced in the next available block of the same epoch, but the nominator keeps their nomination active until the last block of that epoch. At that point, the tokens are returned. The nominator cannot move those tokens before the epoch ends.
-
-**Un-nominate now**
-
-A participant can choose to un-nominate at any time, and the action is executed immediately following the block it is announced in (within the same epoch). 
-
-The participant will not receive any rewards from the validator in that epoch. The tokens are marked as available to the participant.
-
-### Staking rewards
-Validators and nominators both receive incentives from the network, depending on factors including how much stake is nominated. 
-
-**To be considered for staking rewards, a participant must associate VEGA to a Vega key and nominate one or more validators.**
-
-For restricted mainnet, rewards are distributed among validators in proportion to their total stake. The total stake includes a validator's own stake and the tokens nominated to that validator. 
-
-At the end of each epoch, reward payments are calculated per active validator, and then some of that reward is divided between their nominators. The proportion that goes to nominators is defined by the network parameter `reward.staking.delegation.delegatorShare`. 
-
-:::note Further reading
-Read the [staking rewards](https://github.com/vegaprotocol/specs/blob/main/protocol/0058-simple-POS-rewards.md) spec for full details for how rewards are calculated and will be in future iterations. 
-:::
- 
-:::info
-VEGA tokenholders can use [token.vega.xyz](https://token.vega.xyz) to associate their tokens and nominate validators to receive rewards. CoinList
-custodial users should confirm with CoinList how staking works for them.
-:::
-
 
 ### Bridges used for staking
 Because VEGA is an ERC20 token, all actions regarding staking are initiated on Ethereum, rather than on the Vega protocol. This allows VEGA to be staked to a Vega public key without any action on the Vega network, and without putting the tokens under the control of the Vega network.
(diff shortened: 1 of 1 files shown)


docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,75 @@ title: Vega Chain
...
 ---
 
-Vega uses Tendermint as a consensus layer to form a blockchain. The rest of the information here informs on how that blockchain and its relevant components is comprised. [Read more about how Vega bridges to Ethereum](/docs/concepts/vega-chain/#bridges-used-for-staking). 
+Vega uses Tendermint as a consensus layer to form a blockchain. The rest of the information here informs on how that blockchain and its relevant components is comprised. 
+
+Read more: [How Vega bridges to Ethereum](/docs/concepts/vega-chain/#bridges-used-for-staking)
 
 ## Delegated proof of stake
-Vega runs on a delegated proof of stake blockchain. Participants--validators and token-holders--use their VEGA tokens to nominate the validator nodes that run the network. Non-validator participants assign the voting rights of their VEGA tokens to endorse a validator's trustworthiness. 
+Vega runs on a delegated proof of stake blockchain. Participants -- validators and token-holders -- use their VEGA tokens to nominate validator nodes that run the network. Non-validator participants assign the voting rights of their VEGA tokens to endorse a validator's trustworthiness.
```
campbellssource, line 12: suggestion (non-blocking) I wonder if this can do more to explain the slightly novel (to me) way that Vega does POS... that doesn't slash bad validator's stake but instead reduces the rewards they would have received from trading fees that epoch....
Token holders nominate validators they believe to be trust worthy, if a validator turns out to do things badly it (and it's delegates), forfeit staking rewards from trading fees that epoch. Then token holders may change their nomination, resulting in fewer rewards for un-trustworthy validators.

docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,75 @@ title: Vega Chain
...
+Read more: [Penalties](/docs/concepts/vega-chain#penalties)
 
 ### VEGA token
-Vega uses the VEGA ERC20 token for governance, which includes nominating validators, and creating and voting on governance proposals.
+Vega uses the VEGA ERC20 token for governance, which includes nominating validators to run nodes, and creating and voting on governance proposals.
 
 A VEGA token (or fraction) can be either unassociated or associated with a Vega key:
 
 * **Unassociated**: The tokenholder is free to do what they want with the token, but cannot nominate a validator with it
 * **Associated**: The token is locked in the staking smart contract and can be used to nominate a validator. It must be unassociated to withdraw it
```
campbellssource, line 30: suggest: Associated with a Vega key (using an ethereum smart contract) so that it can nominate a validator.

docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,75 @@ title: Vega Chain
...
+
+Read more: [Governance of Vega](/docs/concepts/vega-protocol#governance)
 
 :::info
 A user's VEGA tokens must first be associated with a Vega key before they can be used for governance and staking.
 :::
 
-### Staking on Vega 
-Vega networks use the ERC20 token VEGA for staking. Staking requires the combined action of associating VEGA tokens (or fractions of a token) to the Vega staking bridge contract; and using those token(s) to nominate one or more validators. 
+### Bridges used for staking
```
campbellssource, line 40: suggest this is something more like:
Vega listens to stake events from staking bridges (set by network param). At time of writing there are two, one for general use and one on the Vesting contract that holds tokens that are still locked according to a vesting schedule so that they can Stake and Govern while not being freely traded.

docs/concepts/vega-chain.md
```diff
@@ -85,61 +84,117 @@ Exceptions to automatic nomination:
 ### Un-nominating validators
 Participants can remove their nomination at the end of an epoch, or immediately. The un-nominated tokens will be restored back to the participant's associated token balance.
 
-**Un-nominate towards the end of the epoch** 
-
+#### Un-nominate towards the end of the epoch
 A participant can un-nominate towards the end of the current epoch, which means the stake is not used for the validator from the following epoch. 
 
 The action is announced in the next available block of the same epoch, but the nominator keeps their nomination active until the last block of that epoch. At that point, the tokens are returned. The nominator cannot move those tokens before the epoch ends.
```
campbellssource, line 91: Suggest including: this way a the token holder (and its nominated validator) is still entitled to the rewards from that epoch. using un-delgate now forfeits any rewards from that epoch

docs/concepts/vega-chain.md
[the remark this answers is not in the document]
Vegaklaus: Comment (non-blocking)This is slightly misleading about validator tokens. I would say "Participants (i.e., token-holders) use their Vega tokens....., whereas there is a minimum amount of tokens Validators are required to own and delegate to themselves.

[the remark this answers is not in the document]
Vegaklaus: In above comment, how about: In addition to being delectable to choose validators, tokens are also granting the token holder voting rights on governance decisions and network parameters. If a token is delegated, the voting rights stay with the token owner and are not transferred to the validator in question.

Vegaklaus: Agree with above suggestion."Vega is non slashing, i.e., there is no mechanism through which a token holder can lose a delegated token through a validator being punished; any measures to that end use different mechanism that will hurt a bad validators (and potentially their delegators) revenue, but not the delegated tokens.

docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,75 @@ title: Vega Chain
...
+Read more: [Governance of Vega](/docs/concepts/vega-protocol#governance)
 
 :::info
 A user's VEGA tokens must first be associated with a Vega key before they can be used for governance and staking.
 :::
 
-### Staking on Vega 
-Vega networks use the ERC20 token VEGA for staking. Staking requires the combined action of associating VEGA tokens (or fractions of a token) to the Vega staking bridge contract; and using those token(s) to nominate one or more validators. 
+### Bridges used for staking
+Because VEGA is an ERC20 token, all actions regarding staking are initiated on Ethereum, rather than on the Vega protocol. This allows VEGA to be staked to a Vega public key without any action on the Vega network, and without putting the tokens under the control of the Vega network.
```
Vegaklaus, line 42: Comment (non-blocking): I find this a little misleading; the linking/unlinking a token to a Vega key is initiated by ethereal, the choice of validators is happening on Vega. Also might be useful to add something along the line of "All governance voting and change of nominations happen exclusively on the Vega chain, thus the only Ethereum gas fees that occur are the process of associating tokens and transferring the rewards.

docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,75 @@ title: Vega Chain
...
 
-**Epochs** 
-An epoch is a time period during which staking changes can be announced and then implemented. Changes that are announced in one epoch will only be executed in the following epoch (excepting 'un-nominate now' - see below). The length of an epoch is set by the network parameter `validators.epoch.length`. 
+* Staking **unlocked tokens** is done using the staking bridge contract. The staking bridge contains functions enabling users to deposit, remove, and transfer stake by moving the governance tokens from their wallet to the staking bridge.
```
Vegaklaus, line 44: Comment (non-blocking) I couldn't call it moving the tokens, I interpret it more as notifying the staking bridge what tokens are associated to which key @C0deMunk33

docs/concepts/vega-chain.md
Vegaklaus: Add to above: 'un-delegate now or moving the tokens to a different Ethereum adress'

docs/concepts/vega-chain.md
```diff
@@ -85,61 +84,117 @@ Exceptions to automatic nomination:
...
 
-At the end of each epoch, reward payments are calculated per active validator, and then some of that reward is divided between their nominators. The proportion that goes to nominators is defined by the network parameter `reward.staking.delegation.delegatorShare`. 
+In each epoch, rewards are distributed among validators in proportion to the number of tokens they represent (i.e., their total stake). The total stake includes a validator's own stake and the tokens nominated to that validator. Of this reward, a fixed amount is distributed among the tokenholders the validator represents. Currently that is 88.3%, though validators can vote to change that value.
 
-:::note Further reading
-Read the [staking rewards](https://github.com/vegaprotocol/specs/blob/main/protocol/0058-simple-POS-rewards.md) spec for full details for how rewards are calculated and will be in future iterations. 
-:::
+The reward scheme uses a linear reward curve - the reward per staked token is independent of the behaviour of other tokenholders. This holds for validators as well, with the exception that there is a maximum amount of stake an individual validator can take on. 
+
```
Vegaklaus, line 107: However, to avoid validators getting too big, the reward a validator gets (and can distribute to its delegators) is capped, i.e., the reward per token is decreasing if a validator exceeds a maximum size.

candida-d: Added, thank you!

docs/concepts/vega-chain.md
candida-d: I've added most of this, though in a slightly different form. 

Because the repercussions of 'un-nominate/delegate now' are described in that section, I left that as-is.

docs/concepts/vega-chain.md
candida-d: Updated to replace "moving tokens"!

docs/concepts/vega-chain.md
candida-d: I've included your suggestions, and reorganised the section slightly to (hopefully) remove anything misleading about the eth/vega interaction

docs/concepts/vega-chain.md
candida-d: I've incorporated the ideas from your staking bridge intro, let me know if you think it does the job

docs/concepts/vega-chain.md
candida-d: I believe this is already covered by the line above, and I don't want to define associating by using the word associating :)

docs/concepts/vega-chain.md
candida-d: Non-slashing added!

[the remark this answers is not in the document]
candida-d: I've reorganised this section a bit based on your comments here on dPoS, and hopefully it's more detailed and clearer.

docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,89 @@ title: Vega Chain
...
+Vega uses Tendermint as a consensus layer to form a blockchain. The rest of the information here informs on how that blockchain and its relevant components is comprised. 
+
+Read more: [How Vega bridges to Ethereum](/docs/concepts/vega-chain/#bridges-used-for-staking)
 
 ## Delegated proof of stake
-Vega runs on a delegated proof of stake blockchain. Participants--validators and token-holders--use their VEGA tokens to nominate the validator nodes that run the network. Non-validator participants assign the voting rights of their VEGA tokens to endorse a validator's trustworthiness. 
+Vega runs on a delegated proof of stake blockchain. 
+
+Validator nodes run the Vega network, and they deciding on the validity of the blocks containing the network's transactions and thus execute those transactions. The validators who run validator nodes are required to own a minimum amount of VEGA tokens that they delegate to themselves.
```
ze97286, line 14: they *are* deciding?

candida-d: yes they are haha, thank you

candida-d, line 14:
```suggestion
Validator nodes run the Vega network, and they decide on the validity of the blocks containing the network's transactions and thus execute those transactions. The validators who run validator nodes are required to own a minimum amount of VEGA tokens that they delegate to themselves.
```

docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,89 @@ title: Vega Chain
...
+* Staking **unlocked tokens** uses the staking bridge contract. The staking bridge contains functions enabling users to deposit, remove, and transfer stake by notifying the staking bridge what tokens are associated to which Vega key. 
 
-:::info
-VEGA tokenholders can use [token.vega.xyz](https://token.vega.xyz) to associate their tokens and nominate validators. A Vega Wallet and Ethereum wallet are both required. CoinList custodial users should confirm with CoinList how staking works for them.
+* When staking **locked tokens**, the Vega node interacts with the ERC20 vesting contract, which holds tokens that are locked per a vesting schedule, and provides the same utility as the staking bridge smart contract. This allows locked tokens to be used for staking and governance while not being freely tradeable. 
+
+Whether tokens are unlocked or locked, the bridge events make the Vega network of how many tokens a given party has associated and/or unassociated.
```
ze97286, line 61: make the Vega network _know_ of how many

docs/concepts/vega-chain.md
```diff
@@ -4,76 +4,89 @@ title: Vega Chain
...
+All events (including the above, plus stake per validator and others) are only registered after a certain number of block confirmations, as defined by the network parameter `blockchains.ethereumConfig`. 
+
+:::note Further reading
+**[Staking Bridge contracts](https://github.com/vegaprotocol/Staking_Bridge)** - on Vega's staking bridge GitHub repository.
 :::
 
-### Maximum stake per validator
-Each validator has a maximum amount of stake that they can accept. During restricted mainnet, this will be the same amount for all validators. 
+### Spam protection
+To avoid fragmentation or spam, there is a minimum delegateable stake that defines the smallest unit (fractions of) tokens that can be used for nomination, defined by the network parameter `validators.delegation.minAmount`.
```
ze97286, line 70: minAmount is not a very strong spam protection because the transaction will be processed, it will just be declined. There are better ones employed:
1) you need to have a balance of `spam.protection.delegation.min.tokens` to be able to submit a delegation transaction (I think that's currently 0.1 VEGA on mainnet) 
2) you can't run more than `spam.protection.max.delegations delegation` transactions per day.

docs/concepts/vega-chain.md
```diff
@@ -83,63 +96,123 @@ Exceptions to automatic nomination:
...
+### Rewards
+Validators and nominators both receive incentives for securing the network. The amount of those incentives, rewarded as VEGA, depends on factors including how much stake is nominated. 
 
-For restricted mainnet, rewards are distributed among validators in proportion to their total stake. The total stake includes a validator's own stake and the tokens nominated to that validator. 
+**To be considered for staking rewards, a tokenholder must associate VEGA to a Vega key and nominate one or more validators.**
 
-At the end of each epoch, reward payments are calculated per active validator, and then some of that reward is divided between their nominators. The proportion that goes to nominators is defined by the network parameter `reward.staking.delegation.delegatorShare`. 
+In each epoch, rewards are distributed among validators in proportion to the number of tokens they represent (i.e., their total stake). The total stake includes a validator's own stake and the tokens nominated to that validator. Of this reward, a fixed amount is distributed among the tokenholders the validator represents. Currently that is 88.3%, though validators can vote to change that value.
```
ze97286, line 120: the last statement is a bit inaccurate. this 0.883 is applied to the proportion of delegation by delegators, not on the reward amount that goes to the validator and its delegators. 

Lets say for example that validator 1 is due to get 1000 tokens for them and their delegators. 
And lets say that they have self stake of 3000 and delegation by delegators of 7000. 
That means that the delegators of validator1 will get 0.883 * 1000 * 7000/10000 = 618 which will be divided between the delegators to validator 1 based on their relative proportion of nomination in the validator. 
Validator 1 would get the rest 1000-618 = 382 tokens.

docs/concepts/vega-chain.md
```diff
@@ -83,63 +96,123 @@ Exceptions to automatic nomination:
...
+
+If, at the end of an epoch, a validator does not have sufficient stake self-nominated or has overall too much stake, then their validator score will be lowered, which can impact the rewards a validator and its' nominators receive. 
+
+Below are the two factors that can lower a validator's score, and why. 
+
+#### Not enough self-nominated stake
+Self-nominated stake refers to the amount of VEGA a validator has staked to their own node.  The minimum stake amount required is set using the network parameter `reward.staking.delegation.minimumValidatorStake`. Not having enough self-nominated stake can have an impact on rewards. 
+
+* **Network risk**: A validator who has not committed enough stake to meet the minimum is a risk to the network because they may not be invested in keeping the network running
+* **Validator score**: If a validator does not meet the `reward.staking.delegation.minimumValidatorStake`, then the validator's score is set to zero
```
ze97286, line 167: that's not yet true in 0.47.6. If that was correct neither the validator nor parties who nominated them wouldn't get a reward which is still not the case. The only sanction is that the validator itself isn't rewarded, but it's score is what it is based on its total stake.

docs/concepts/vega-chain.md
candida-d, line 61:
```suggestion
Whether tokens are unlocked or locked, the bridge events let the Vega network know of how many tokens a given party has associated and/or unassociated.
```

docs/concepts/vega-chain.md
candida-d, line 167:
```suggestion
* **Validator score**: If a validator does not meet the `reward.staking.delegation.minimumValidatorStake`, the validator's score is lowered, which can affect rewards
```

candida-d: Ah, I see what you mean, both can't be true at the same time. I've changed to make it less specific

docs/concepts/vega-chain.md
candida-d, line 70:
```suggestion
There are several spam protections enabled to protect the Vega network. 

- A participant who wants to submit a delegation (nomination) transaction, needs to have a balance of at least the minimum defined by the network parameter `spam.protection.delegation.min.tokens` to be able to submit the transaction. (Currently set as 0.1 VEGA)
- A participant cannot do more delegation transactions than the max set by `spam.protection.max.delegations` per day. (Currently set to 360.)
```

docs/concepts/vega-chain.md
candida-d, line 120:
```suggestion
In each epoch, rewards are distributed among validators in proportion to the number of tokens they represent (i.e., their total stake). The total stake includes a validator's own stake and the tokens nominated to that validator. Of this reward, a fixed amount is distributed among the tokenholders the validator represents.
```

candida-d: I've cut that sentence for now but will reflect this better in another PR, thanks for explaining it!

PR #60, opened 2022-03-18
Rewards & Updating desktop wallet

candida-d:
- Make explicit that staking rewards need to be withdrawn before they can be staked 
- Add instructions for updating the desktop wallet app


campbellssource:
LGTM

PR #67, opened 2022-04-05
DO NOT MERGE Add trading concepts and data feeds

candida-d:


edd:
This PR was created before #84 and so it's hard to bring up to date. #113 is a less subtle attempt to do that by copy and pasting from this branch.

PR #97, opened 2022-05-05
Fix trailing slashes breaking relative links

edd:
@jeremyletang found an interesting case where a link is broken only if you navigate directly to a page. As per Docusaurus issue 3372, I have:
- Updated the Netlify setting to not mess with the URLS

This PR follows that up by adding another config option that explicitly should make sure that trailing slashes aren't ever generated


edd:
@candida-d I believe this is ready for review now. Links appear to work consistently across visiting a page and navigating to the page.

I also updated Docusaurus and some plugins as part of this.

candida-d (approved):

PR #128, opened 2022-06-28
Add governance concepts for testnet docs version

candida-d:


docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+
+When creating a market governance proposal, whether it is for a new market or to change parameters for an existing market, it's recommended that you sense check the proposal and share the final details with the tokenholder community before proposing, so that you can garner support and make any necessary amends. 
+
+Read more:
+* [Vega community forum](https://community.vega.xyz): Share your draft proposals for community discussion.
+* [New market proposal](../tutorials/proposals/new-market-proposal.md): Guide to submitting a proposal for a new market using the command line
+* [Update market proposal](../tutorials/proposals/update-market-proposal.md): Guide to submitting a proposal to change a market using the command line
+
+### Propose a new market
+Tokenholders can propose new markets, which then need to be voted on by other tokenholders. If the market proposal gets a majority of tokeholder support, then it will be enacted. The required majority is defined by the network parameter `governance.proposal.market.requiredMajority`.
```
barnabee, line 92: I think there are similar net params for market governance, there are also various other net params that could be referenced for a number of the comments mentioned on this page. Should we include their details?

barnabee: Bonus points for adding a component that shows the value of a network param

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+* [Vega community forum](https://community.vega.xyz): Share your draft proposals for community discussion.
+* [New market proposal](../tutorials/proposals/new-market-proposal.md): Guide to submitting a proposal for a new market using the command line
+* [Update market proposal](../tutorials/proposals/update-market-proposal.md): Guide to submitting a proposal to change a market using the command line
+
+### Propose a new market
+Tokenholders can propose new markets, which then need to be voted on by other tokenholders. If the market proposal gets a majority of tokeholder support, then it will be enacted. The required majority is defined by the network parameter `governance.proposal.market.requiredMajority`. 
+
 :::info
-You'll need a Vega Wallet for staking and receiving rewards. Connect to wallets and see your account balance on the [Vega token website](https://token.vega.xyz). CoinList custodial users should confirm with CoinList how staking works for them.
+A liquidity commitment is optional when proposing a market, but a market will not enter into continuous trading until its liquidity needs are met.
```
barnabee, line 95: Is it optional or no longer allowed (or going to be no longer allowed)? I thought we decided to remove the ability to submit them at the same time entirely but may be wrong?

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+* Risk model parameters
+* Product specifics including the settlement asset and quote name
+* Decimal places for the settlement asset, market, and positions 
+* Oracle details, including the oracle's public key, specifications for the settlement price and trading termination, and data filters
+* Liquidity monitoring parameters, including the target stake parameters, triggering ratio and auction extension
+
+Optional fields include: 
+* Metadata so that people can easily interpret the market's details
+* Price monitoring parameters, including the triggers covering the horizon, probability and auction extension time. If left blank these parameters will default to the values set in the network parameters
+* Liquidity commitment: the amount committed, proposed fee level, and the buy and sell order shapes. Note: Once a market is proposed, it can accept liquidity commitments from any party
```
barnabee, line 111: See above - is this no longer going to be allowed?

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+Required fields include: 
+* Instrument details, including a human-readable name, an understandable shortcode for the market, the type of product (futures)
+* Risk model parameters
+* Product specifics including the settlement asset and quote name
+* Decimal places for the settlement asset, market, and positions 
+* Oracle details, including the oracle's public key, specifications for the settlement price and trading termination, and data filters
+* Liquidity monitoring parameters, including the target stake parameters, triggering ratio and auction extension
+
+Optional fields include: 
+* Metadata so that people can easily interpret the market's details
```
barnabee, line 109: It is optional but we should strongly encourage this and also somehow have a resource to discover convetions for various type of standard and useful metadata. Not sure the best way to achieve this… maybe community team have been thinking about this (or should be) when they are thinking about the whole process...

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+<!--
+* [Data sources]
+* [Liquidity monitoring parameters]
+* [Price monitoring parameters]
+-->
+
+### Risk models and parameters
+When proposing a market, the market proposer will need to choose the risk parameters associated with the risk model that's appropriate for the instrument. The risk model is essential for calculating margins on the market. 
+
+The first product available to create is cash-settled futures, which uses a log-normal risk model. While the model is pre-defined, you'll need to choose the individual parameters.
```
barnabee, line 126: > cash-settled futures, which uses a log-normal risk model

"which use" though actually it is more that log normal is the only risk model we currently support than that cash settled futures need to or will always only use that model

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+
+You should choose parameters that ensure the risk model adequately represents the dynamics of the underlying instrument, and that the resulting margins strike the right balance between prudence and capital efficiency.
```
barnabee, line 128: We should ask David or Witold to create a guide for how to work these out and perhaps a spreadsheet template and/or some Python script to help (and link it/them here)

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+  * suggested value: `0`.
+
+The remaining, model specific parameters are covered below.
+
+:::note Further reading
+**[Margins and Credit Risk on Vega](https://vega.xyz/papers/margins-and-credit-risk.pdf)** - Note, a position size of 1 is assumed throughout the research paper.
 :::
 
+#### Log-normal risk model
+The log-normal model assumes that the price of the underlying asset follows the process specified by the stochastic differential equation:
```
barnabee, line 151: This is definitely a change of pace/style 🙃

barnabee: More seriously I wonder if there's a layman's explanation that would be useful

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+The first assets that will be available for interacting with markets on Vega will be ERC20 assets. They will need to be deposited into the ERC20 bridge contract. The funds in that smart contract will then be made available to the user's chosen Vega public key.
+
+:::info 
+Associated and deposited are not equivalent, as deposited tokens are held within the ERC20 bridge contract, and associated tokens stay in an Ethereum wallet or in the vesting contract.
+:::
 
-The first assets that will be available for interacting with markets on Vega will be ERC20 assets. They will need to be deposited into the ERC20 bridge contract. The funds in that smart contract will then be made available to the user's chosen public key.
+To acquire an asset balance on the Vega network, a participant must first deposit an asset using the relevant Vega asset bridge. Every asset class supported by and voted into Vega will have a bridge.
```
barnabee, line 216: I'd say type of asset rather than asset class

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+A proposal to change the `feeAmtValidators.MinimumProposalPeriod` would need to pass all of the thresholds listed above.
+
+<!--### Threshold and rules [WIP]-->
+  
+## Asset management
+Assets used for trading, paying fees, funding rewards, and providing liquidity need to be deposited using a bridge contract, and can be withdrawn back into an Ethereum wallet if they are not being used for margin or liquidity commitment.
+
 ### Deposits
-For restricted mainnet, the deposits function for Vega is not required. To stake, instead of depositing, tokens must be associated to a Vega key. Tokens used for staking stay in your Ethereum wallet, rather than being held in the ERC20 bridge contract. [Read more about staking VEGA tokens.](./vega-chain#staking-on-vega)
+The first assets that will be available for interacting with markets on Vega will be ERC20 assets. They will need to be deposited into the ERC20 bridge contract. The funds in that smart contract will then be made available to the user's chosen Vega public key.
```
barnabee, line 210: Should make clear that each asset (ERC20 token contract) needs to be added the bridge before it will be supported.

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
-Note: Associated and deposited are not equivalent, as deposited tokens are held within the ERC20 bridge contract, and associated tokens stay in an Ethereum wallet or in the vesting contract.
+Once a successful withdrawal transaction has occurred, the ERC20 bridge will emit an `Asset_Withdrawn` event, and confirms to the Vega network that the withdrawal has been completed.
+
+**Read more**: [ERC20 bridge logic API documentation](../api/bridge/contracts/ERC20_Bridge_Logic.md#withdraw_asset)
+
+### Diagram: Withdrawals
+![Withdrawal diagram](/img/concept-diagrams/diagram-withdraw.png)
+
+##### Withdrawing VEGA
+VEGA (an ERC20 token) used for staking is associated with a Vega key, and so to withdraw those tokens, they must be dissociated first.
```
barnabee, line 261: Do you mean withdraw as in withdraw from bridge here? If not is the double meaning of withdraw confusing?

docs/tutorials/proposals/new-market-proposal.md
```diff
@@ -67,9 +67,9 @@ Decimal places need to be defined for both order sizes and the market.
 * `decimalPlaces` - sets the smallest price increment on the book
 * `positionDecimalPlaces` - sets how big the smallest order / position on the market can be
 
-Timestamps are required for ending the voting period, as well as enacting the changes.
+Timestamps are required for ending the voting period, as well as enacting the market. The time between closing and enactment also defines how long an opening auction will be. It cannot be longer than 7 days.
```
barnabee, line 70: Is it 7 days or some network parameter that is initially 7 days

docs/concepts/vega-protocol.md
```diff
@@ -2,69 +2,273 @@
...
+* The governance proposal is checked and then accepted by the wallet as a transaction.
+* The validator nodes then check and validate the proposal. This is when the proposal data that defines the minimum duration, minimum time to enactment, minimum participation rate, and required majority are evaluated against the network's requirements. If not specified on the proposal, the required participation rate and majority for success are defined and copied to the proposal.
+* If the above conditions are not met, the proposal will be rejected and will not be available for a vote. **You'll need to fix and re-submit the proposal.**
+
+:::info 
+Read the [creating and submitting proposals guides](../tutorials/proposals/) to understand how tokenholders can submit proposals using the command line. 
+:::
+
+#### 4. Voting 
+VEGA tokenholders can vote for or against any active proposals, as long as they tokens they want to vote with are associated with their key. It's not necessary to nominate validators with all the tokens, but the tokens must be associated to the Vega key used for voting.
```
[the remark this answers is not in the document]
candida-d:
```suggestion
VEGA tokenholders can vote for or against any active proposals, as long as they tokens they want to vote with are associated with their key. It's not necessary to nominate validators, but the tokens must be associated to the Vega key used for voting. 
```

docs/tutorials/proposals/new-market-proposal.md
edd: It’s the difference between two network parameters, maxClose and maxEnact

docs/concepts/vega-protocol.md
candida-d: It looks like it's actually optional rather than disallowed: https://github.com/vegaprotocol/vega/blob/d92d87a37536a0775652bb0b978792a89316197e/governance/engine.go#L490

PR #141, opened 2022-07-12
Add info on how to handle offline nodes for desktop wallet

candida-d:


daunatv (approved):
As per the testing this should be sufficient for users to update their networks and use the wallet without crash.

PR #148, opened 2022-07-15
Feat/network param embed

### LINKED ISSUE #147: Cut python requirement from build.sh
The very handy script for fetching protos/graphQL schema from private repos is no longer required, as they are public. As such the python setup can be retired in favour of a one line curl/wget

# Tasks
- [ ] Remove `github_get_file`
- [ ] Remove venv setup and related from `build.sh`
- [ ] Remove requirements.txt
- [ ] Replace with CURL

### LINKED ISSUE #158: Make a component to show network parameters and their current value
Some documentation needs to refer to configuration that is set by the network parameters, which means they can be changed by governance. We don't want to add values that will go out of date, so instead let's make a component that can show the currently configured value.

# Tasks (v1)
- [x] Create a react componenet that can display a single network paramter
- [x] Make it able to show just the network parameter name, the name and value, or just the value
- [x] Make it link to the Block Explorer > Network Parameters page so the user can see how the value was set

# Follow up
#157

edd:
Create a component that displays network parameters for a network in the body of a page. Uses some fetched at build time as defaults, then fetches live ones. Closes #158 

Also closes #147, because I was in the build file anyway.

# Preview
(See https://deploy-preview-148--vega-docusaurus.netlify.app/docs/testnet/tutorials/proposals/new-market-proposal)
Screenshot 2022-07-25 at 15 52 07

## Follow up work
My intention creating this was to be able to display network parameters 'live', but in the body of documentation. I achieved this on pages written with it in mind, but going and retrofitting it in to other documents it's not so stellar. See #157 

The component also does no formatting of the values. For example some of the VEGA token values are to 18 decimal places, but in place would look rubbish.

# Todo
- [x] `tokens` should be `token` if it's `1`
- [x] Multiple instances on one page = multiple fetches. Add a singleton store.
- [x] @campbellssource Choose a better emoji
- [x] Create a follow up task to add this component to all docs that use network parameters #157


campbellssource:
I think the 👀 `:eyes:` emoji is probably better. I like that clicking it takes you to the block explorer, I think this adds to me wanting to use eyes.
close second option was 🔍 
then 🌐

edd:
> I think the 👀 `:eyes:` emoji is probably better. 

Done

docs/tutorials/proposals/freeform-proposal.md
```diff
@@ -20,6 +21,10 @@ import TabItem from '@theme/TabItem';
 # Create a freeform proposal
 The aim of this is to allow community to provide votes on proposals which don't change any of the behaviour of the currently running Vega blockchain. That is to say, at enactment time, no changes are effected on the system, but the record of how token holders voted will be stored on chain. 
 
+Note that to create a freeform proposal you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minProposerBalance" hideName={true} suffix="tokens" />
+, while to vote you only need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minVoterBalance" suffix="tokens" hideName={true}/>.
```
candida-d, line 25:
```suggestion
, while to vote you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minVoterBalance" suffix="tokens" hideName={true}/>.
```

candida-d (approved):
Cool feature! I think we might find ways to improve it (or slightly change it) the more it gets used but this is a good version to start with.

scripts/build.sh
```diff
@@ -10,68 +10,39 @@
...
 
+testnet_network_parameters=https://lb.testnet.vega.xyz/network/parameters
+mainnet_network_paramters=https://api.token.vega.xyz/network/parameters
+
 set -e
 
-grpc_doc_owner=vegaprotocol
-grpc_doc_repo=protos
-grpc_doc_branch="v0.53.0"
+doc_version=v0.52.0
```
candida-d, line 18: Why is this now 0.52?

edd: Bad merge

edd: Fixed!

PR #177, opened 2022-08-02
Expand use of NetworkParameter component

edd:
Use NetworkParameter component in concepts pages

PR #148 added a network parameter component to display current values. This branch replaces all references to network
parameters with the component, and adds configuration for how the name or value are displayed

### DIFF
--- a/versioned_docs/version-v0.53/tutorials/proposals/freeform-proposal.md
+++ b/versioned_docs/version-v0.53/tutorials/proposals/freeform-proposal.md
@@ -21,8 +21,8 @@ import TabItem from '@theme/TabItem';
 # Create a freeform proposal
 The aim of this is to allow community to provide votes on proposals which don't change any of the behaviour of the currently running Vega blockchain. That is to say, at enactment time, no changes are effected on the system, but the record of how token holders voted will be stored on chain. 
 
-Note that to create a freeform proposal you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minProposerBalance" hideName={true} suffix="tokens" />
-, while to vote you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minVoterBalance" suffix="tokens" hideName={true}/>.
+Note that to create a freeform proposal you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minProposerBalance"  formatter='governanceToken' hideName={true} suffix="tokens" />
+, while to vote you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minVoterBalance" suffix="tokens" formatter='governanceToken' hideName={true}/>.
 
 
 <Tabs groupId="newFreeform">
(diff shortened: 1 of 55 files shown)


candida-d:
If this component could allow for formatting based on what the number's actual value is (with decimal points, or as a percentage, for example) that would be grand! 

One use case: reward.staking.delegation.minimumValidatorStake displays as 3000000000000000000000 tokens, but ina formatted number it's actually 3,000.00(etc)

edd:
Screenshots showing off some formatters:

Screenshot 2022-08-25 at 17 01 16

Screenshot 2022-08-25 at 17 02 15


`governanceToken` and `percentage` can be set by passing in `formatter`, so for instance the required participation above is using `percentage`, turning 0.55 in to 55%, and token values in to '1 token', '0 tokens', '300 tokens' from their 18 decimal place equivalents.

versioned_docs/version-v0.53/tutorials/proposals/freeform-proposal.md
```diff
@@ -21,8 +21,8 @@ import TabItem from '@theme/TabItem';
 # Create a freeform proposal
 The aim of this is to allow community to provide votes on proposals which don't change any of the behaviour of the currently running Vega blockchain. That is to say, at enactment time, no changes are effected on the system, but the record of how token holders voted will be stored on chain. 
 
-Note that to create a freeform proposal you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minProposerBalance" hideName={true} suffix="tokens" />
-, while to vote you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minVoterBalance" suffix="tokens" hideName={true}/>.
+Note that to create a freeform proposal you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minProposerBalance"  formatter='governanceToken' hideName={true} suffix="tokens" />
+, while to vote you need <NetworkParameter frontMatter={frontMatter} param="governance.proposal.freeform.minVoterBalance" suffix="tokens" formatter='governanceToken' hideName={true}/>.
```
candida-d, line 25: It looks like the formatter for gov token isn't working as it shows as 0, rather than "anything more than 0" in this case

candida-d (changes requested):
Looking good! The expansion of what it can do is great. I left some suggestions, and spotted that one of the formatting options isn't working. 

Also, what would you think of toning down the colour of the component so it's not so eye-catching?

versioned_docs/version-v0.53/tutorials/proposals/freeform-proposal.md
edd: Fixed. The code originall was like IF this is a token value AND the token value is 0.00000000...1, THEN display 'more than 0'. This network parameter is *actually* 0. So now there's another special case that ends up displaying '0 or more' versus 'more than 0' versus '0.000002' or whatever

edd:
> Also, what would you think of toning down the colour of the component so it's not so eye-catching?

Done

candida-d (approved):

PR #199, opened 2022-08-16
Wayfinding

neildawson:
Exploration of how we can help visitors understand the scope of what they can do with Vega today, and signpost key resources.

* Landing page:
  * Added CTA to the hero area so there's a path of least resistance to the concepts area.
  * Replaced the default "features" blocks with key topics, including links to key resources for each.
  * Added a "Spotlight" section where miscellaneous resources can be highlighted.
* On the Vega Protocol (in the concepts section) and governance proposals index pages, added a "topic: governance" element to link these related areas together. I've been referring to this element as a "stub".

All content is placeholder.


neildawson:
I had a chat with Oli to determine if we could use the existing Docusaurus styles for the cards on the landing page. It's not straightforward, and Oli thinks we may have to build a custom theme to do this properly. In the meantime I found the source CSS rules and copy and pasted them into our custom.css file, and updated the spacing. As a result, the cards look and behave almost exactly like the 'native' ones. The only differences: we're not truncating the descriptions or adding an emoji to the title.

I also added a wrapper to the topic stubs so they can be targeted for styling. I think this is a bit easier to maintain than swizzling the details component, but it's open to discussion!

Issue #179, opened 2022-08-04
Update documentation for wallet 0.54


edd:
See: https://github.com/vegaprotocol/vega/issues/5766

# tl;dr
- [vegaprotocol/go-wallet](https://github.com/vegaprotocol/go-wallet) is being deprecated. 
- https://github.com/vegaprotocol/vega/issues/5766 will push people to upgrade at the appropriate time
- The wallet that works for 0.54 will not work with networks below 0.54
- The wallet that works for older networks will not work on 0.54

# Tasks
- [x] ~Update all commands to the new format (probably `vega wallet` instead of `vegawallet`~ Not required
- [x] Prune old wallet documentation (may need to be a separate issue so that it happens at a more appropriate time
- [x] Update tools links to point to the proper repo

edd:
Created for future reference, marked as blocked while 0.54 still isn't out

PR #194, opened 2022-08-15
chore: update release notes to support 0.54.0

gordsport:
Prepares docs for a testnet release of 0.54

- Adds release notes for the changes included in 0.54.0 due to go to testnet imminently
- Update proposal samples for v0.54
- Update GRPC docs for v0.54
- Update REST docs for v0.54
- Update GraphQL docs for v0.54
- Add API descriptor files to `data/v0.54` - this should be for this release only, assuming https://github.com/vegaprotocol/vega/issues/5670 is complete

Closes #186 . Closes #153 .

PR #203, opened 2022-08-18
Add docs for new wallet api

candida-d:
Add a new API section for Vega Wallet as of 0.54

- Add overview
- Add OpenRPC page (generated by [vegaprotocol/openrpc-md](https://github.com/vegaprotocol/openrpc-md))
- Add a page with an embedded [Open RPC playground ](https://github.com/open-rpc/playground)

Undrafted now #194 is merged


mattrussell36:
@candida-d Heres some feedback on the docs.

- Getting started 
  - In the top section I think the structure could be better. Rather than having the green and yellow callouts I think some regular sub headings, one for JSON RPC (and a link to the docs), and one for software compatibility would be easier to read.
  - In the bottom section I'd leave out the bullet for method, params, id. Its all in the json rpc docs and examples.
- OpenRPC Wallet API
  - I think it would be good to see all available methods in a concise list, a contents section if you will, which link to the detailed ection for the method. I realise this is in the playground too but I think it would be good here. Its probably the first thing a lot of people will look for and the playground can be slow to load.
  - The tables for Parameters and Result have a description column which is often empty. Would be good to fill these in if possible or remove?
  - Just a personal visual thing for me but the code block in the result title makes things look too busy.
 Screen Shot 2022-08-20 at 11 39 08 PM
  - I find the top section a little busy. IMO it would be cleaner to remove the green callout for the playground and move the yellow callout about versioning into its place. 
Screen Shot 2022-08-20 at 11 41 35 PM
  - Do we need the quote block here? It kind of just repeats the same info and makes things a little cluttered
Screen Shot 2022-08-20 at 11 43 18 PM

edd:
I've addressed lots of @mattrussell36's feedback in a newly regenerated page:
- Green header is gone for now. We added it manually, so regenerating just... removed it. I think that's OK now that it's in the sidebar, which it wasn't when we added the header
- Added backticks around the method headers, which makes them stand out better
- Added a TOC at the start that contains the name and the summary and links to the method section
- Removed the summary underneath the method header, which helps reduce the over-formatting
- ENUM params are now formatted better

mattrussell36 (approved):

mattrussell36:
A couple other things.

- For whatever reason I found it hard to find what the actual jsonrpc endpoint was. I feel like this should be more prominent somewhere. Perhaps a full section on getting started?
- Typo in session.request_permissions in the reference page. The permission params should be `public_keys` not `public_key`

candida-d:
@ValentinTrinque I think this one is for you: 

> Typo in session.request_permissions in the reference page. The permission params should be public_keys not public_key

candida-d:
> 
>     * For whatever reason I found it hard to find what the actual jsonrpc endpoint was. I feel like this should be more prominent somewhere. Perhaps a full section on getting started?

@mattrussell36 I've added the endpoint to the top of the page, where it already says what kind of API it is. Is that clearer now?

edd:
I *think* this closes https://github.com/vegaprotocol/documentation/issues/179, assuming the tools links are updated

PR #208, opened 2022-08-23
Improve indexes

neildawson:
### Changes
* Added index pages for most sidebar categories. Some are markdown files, some are autogenerated, see notes.
* Added descriptions to some of their child pages, so the index cards are more meaningful.

### Benefits
* Clicking the category in the sidebar takes you somewhere, instead of just opening the sidebar group. Some categories already did this, so this is a move towards consistency in sidebar behaviour. Note this commit doesn't make all categories behave in that way, see exceptions in the notes below.
* Visiting a category URL now takes you to a page instead of a 404 error.
* The categories are now clickable in the breadcrumbs on all child pages.
* All indexes will automatically update when child pages are changed.
* The index cards are more scannable than links in lists or tables, and they break up the monotony of big blocks of text on other pages.

### Notes
* The commits so far don't alter any URLs, only change or create pages at existing URLs, so won't break any inward links.
* The "Tutorials" section index is autogenerated. In all other instances I manually added an `index.md` file to the folder as I don't see how to autogenerate them without adding a lot of length to `sidebars.js`. However, this does allow us to add richer content to the indexes, as autogenerated indexes can only have some introductory text.
* Exception: [The "Ethereum bridge" category](https://docs.vega.xyz/docs/testnet/api/bridge) _should_ already work like this, as the "Overview" page has the filename `index.md`, following [the Docusaurus naming convention](https://docusaurus.io/docs/next/sidebar/autogenerated#category-index-convention). It's unclear to me why it doesn't.
* Exception: I tried to get the REST category to behave like this. (Note that this would change the page's URL.) I changed `overview.md` to `index.md`, and updated the URL in `sidebars.js` to point at the new expected URL. I get the following error in terminal, which is confusing, because I can visit the page directly at the same URL. Terminal also spits out a list of possible targets to try instead, including `api/rest/index`, which doesn't work either 🤔

```
[ERROR] Loading of version failed for version current
[ERROR] Error: Invalid sidebar file at "sidebars.js".
These sidebar document ids do not exist:
- api/rest
```

* I also tried doing this for the API section itself, as I believe this would result in the API section being visible in breadcrumbs, which it currently isn't. I had the same problem as described above.


neildawson:
Picking up this PR again after a few conversations, current status:

* @edd knows how to correct the behaviour of the REST and Ethereum bridge categories and can do so after this PR has been accepted.
* @candida-d asked if we could use the autogenerated card index on Node Operators page, but it doesn't seem possible. Trying to do so results in the error: "This page crashed. /docs/testnet/node-operators is not associated with a category. useCurrentSidebarCategory() should only be used on category index pages." It seems Docusaurus is quite strict about where these can be used.
* Added indexes for the new wallet categories in the API area. Again, simply having an "index.md" file does not seem to be sufficient, which I don't understand. In the meantime, I've linked the "Wallet API" category to the intro page.
* No progress on making the API index linkable in breadcrumbs, possibly has the same cause as the Node Operators index above.

edd:
Data node index file: I can't work out exactly why adding the DocCardList doesn't work as expected. The error is that the index doesn't have a category assoiciated, but I'm not sure *why*. It's not because it's autogenerated, and even if I manually add it to the category explicitly, it doesn't work. However!

I think if we just remove the index file now and have it totally autogenerated, it might work. Let's try that in another PR. I don't think that blocks this, because the manually populated list *works*, it's just inconsistent.

edd (approved):

PR #238, opened 2022-09-08
Data sources rework

edd:
While this will be slightly outdated when https://github.com/vegaprotocol/vega/pull/5980 is merged, it reworks the Data Sources 'Trading concepts' page, and brings the Using Oracles  page in line (mainly by renaming it to Using Data Sources.


edd:
Some of this work exists in https://github.com/vegaprotocol/documentation/tree/spec-update-v0.61.0 and will be pulled across to the v0.62.0 branch when it exists

candida-d (commented):
noticed a few small things

docs/concepts/trading-on-vega/data-sources.md
```diff
@@ -1,103 +1,113 @@
...
-
-## Signed message data sources
-Signed message data sources are the first external data source to be supported on Vega, and can be nominated for settling a market. They introduce a Vega transaction that represents a data result. The result is validated by ensuring the signed message is signed by one of a set of public keys provided in a market’s proposal.
-
-:::info
-There are no rewards associated with being a signed message data source, nor are there fees associated with being or using a signed message data source.
-:::
+Specifically, Signed Message Data Sources are equivalent to Posters in [Compound’s Open Price Feed](https://medium.com/compound-finance/announcing-compound-open-oracle-development-cff36f06aad3), taking signed price reports and posting them to the Vega chain. As Open Oracle reports include signatures, the data can still be verified against its source.
```
barnabee, line 33: This is not quite true: the signed message source is the public key and the specification of the fields and filters to process the incoming data, so it is more like the Reporter. Whoever submits the data for the signed message source to Vega is the equivalent of the Poster, but I'm not sure we've named them really.

So in the example the Compound price feed is the signed message data source, the "Reporter" is whoever does the signing, and anyone who submits that txn to Vega is the "Poster"

docs/concepts/trading-on-vega/data-sources.md
```diff
@@ -1,103 +1,113 @@
...
-
-## Signed message data sources
-Signed message data sources are the first external data source to be supported on Vega, and can be nominated for settling a market. They introduce a Vega transaction that represents a data result. The result is validated by ensuring the signed message is signed by one of a set of public keys provided in a market’s proposal.
-
-:::info
-There are no rewards associated with being a signed message data source, nor are there fees associated with being or using a signed message data source.
-:::
+Specifically, Signed Message Data Sources are equivalent to Posters in [Compound’s Open Price Feed](https://medium.com/compound-finance/announcing-compound-open-oracle-development-cff36f06aad3), taking signed price reports and posting them to the Vega chain. As Open Oracle reports include signatures, the data can still be verified against its source. 
 
 A signed message data source must include:
```
barnabee, line 35: We should be clear here whether we mean:

1. A signed message data source definition / specification (i.e. the data included in the market proposal)
2. The data source / feed itself (i.e. the actual data message(s))

This looks like it covers both as if they are the same thing (the pubkeys and data type are from 1 if I understand the below correctly and the ABI encoded data is from 2).

docs/concepts/trading-on-vega/data-sources.md
```diff
@@ -1,103 +1,113 @@
...
+[
+	{
+		"name": "vega.builtin.timestamp",
+		"value": "1661273220"
+	}
+]
+```
+
+## Signed JSON message
+Signed JSON messages are an alternative to using *signed message data sources*, which are best used for off-chain prices. Some markets need data other than prices, and for that flexibility Vega also supports arbitrary JSON messages, signed by a Vega key. Data from these types is only as reliable as the Vega public key that publishes the data, whereas signed message data sources have signed data. The advantage is that the format is less rigid than signed messages: any properties can be specified, including strings, booleans or non-price numbers.
```
barnabee, line 80: Signed JSON messages can and should also be used for prices if they don't come from an Open Oracle source, as they are easier to work with. So anyone creating an oracle afresh just for Vega would probably use these for their price source.

docs/tutorials/using-data-sources.md
```diff
@@ -361,7 +356,7 @@ Vega provides a timestamp source, which is useful for terminating a market at a
 
 As the name implies, a built in data source is generated inside Vega, and cannot be submitted by other keys.
 
-### Using built-in oracle data for trading termination
+### Using built-in data for trading termination
```
barnabee, line 359: We should not that if using time we should always use greater than or equals rather than a straight equals, to avoid missing the time because no event is emitted with the precise required trimestamp.

barnabee (changes requested):
Sorry!

docs/tutorials/using-data-sources.md
```diff
@@ -70,16 +62,19 @@ The following oracle spec would make the market use the BTC value from the Open
...
 ```
 
+The pubKey in this case is the **Ethereum public key that signed the data in the message**. 
+
 ### Submitting Open Oracle data
+
 Use the command line to submit an Open Oracle message as a transaction that is signed by your Vega wallet and sent to the validators for consensus.
 
-Below, find instructions on how to submit Open Oracle data. Anyone can submit Open Oracle data, at any time. Markets should be configured to only use the data at the relevant time, such as after a defined settlement date.
+Below, find instructions on how to submit Open Oracle data as a Signed Message. Markets should be configured to only use the data at the relevant time, such as after a defined settlement date.
```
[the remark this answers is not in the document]
candida-d: notes: the market definition (product itself) needs to include the date filter. the data source defines that only valid data can come through. (make it more explicit)

docs/tutorials/using-data-sources.md
```diff
@@ -244,7 +239,7 @@ Use the command line to submit a JSON message as a transaction that is signed by
 :::
 
 ### 1. Define your JSON structure
-JSON oracles can contain arbitrary JSON data - but to be useful for market creators, they'll need to know what the structure is ahead of time. Pick a data model and ensure that it's well communicated. For this tutorial, we'll create a JSON data source for the number of humans that have walked on the moon:
+JSON data has no defined structure - but to be useful for market creators, they'll need to know what the structure is ahead of time. Pick a data model and ensure that it's well communicated. For this tutorial, we'll create a JSON data source for the number of humans that have walked on the moon:
```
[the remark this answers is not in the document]
edd:
```suggestion
JSON data should be submitted as a single object of attributes and primitive values (i.e. no objects or arrays). Exactly what the attributes are called is up to the submitter of the data. Pick your structure in advance and ensure that it's well communicated. For this tutorial, we'll create a JSON data source for the number of humans that have walked on the moon:
```

edd (commented):
Added a suggestion

docs/concepts/trading-on-vega/data-sources.md
edd:
```suggestion
Specifically, the signer of the Signed Message Data Sources is equivalent to the Reporter in [Compound’s Open Price Feed](https://medium.com/compound-finance/announcing-compound-open-oracle-development-cff36f06aad3). As Open Oracle reports include signatures, the data can still be verified against its source. The Poster equivalent is the Vega key that submits the Signed Message to the Vega chain when the market needs it.
```

candida-d:
Merged in https://github.com/vegaprotocol/documentation/pull/429

PR #244, opened 2022-09-13
Fix new market example

edd:
The asset this market is using is 5 decimal places, while the example uses 18. 

- Fix generated newMarket sample to use 5 decimal places


edd:
Merging without a review. Cowboy 🤠 Verified with @sam-keen .

PR #274, opened 2022-09-21
Network param instances

neildawson:
Embedding the network parameter component in places where they were mentioned, or where they may be useful.

I think we would benefit from a style guide for how we use these, as there are a few patterns at the moment. Maybe that's fine though!


edd (commented):
I fixed the usage of a parameter that doesn't exist and changed the text to make it make sense. @candida-d over to you.

I'd be happy with a style guide/limiting the configurability of the component to Do The Right Thing. I think it's used in enough places that it would make sense to have some rules, so far I have more molded it to the text that exists rather than thinking about consistency - happy to change.

docs/concepts/trading-on-vega/fees-rewards.md
```diff
@@ -45,14 +47,14 @@ At a high level, the trading fee is calculated using the following formulas:
 * Trade value and fee factor: *USDC5000 x 0.003 = USDC15*
 * The fee is the same regardless of the number of transactions the order needs to be completely filled, as long as they trade at the same price.
 
-The fee factors are set through the following network parameters: `market.fee.factors.infrastructureFee`, `market.fee.factors.makerFee`, `market.fee.factors.liquidityFee`.
+Two of the three fee factors are set through network parameters: <NetworkParameter frontMatter={frontMatter} param="market.fee.factors.infrastructureFee" />, <NetworkParameter frontMatter={frontMatter} param="market.fee.factors.makerFee" />. The liquidity fee is set by the liquidity providers on the maret.
```
candida-d, line 50:
```suggestion
Two of the three fee factors are set through network parameters: <NetworkParameter frontMatter={frontMatter} param="market.fee.factors.infrastructureFee" />, <NetworkParameter frontMatter={frontMatter} param="market.fee.factors.makerFee" />. The liquidity fee is set by the liquidity providers on the market.
```

neildawson:
> I'd be happy with a style guide/limiting the configurability of the component to Do The Right Thing. I think it's used in enough places that it would make sense to have some rules, so far I have more molded it to the text that exists rather than thinking about consistency - happy to change.

I agree @edd! I think we have enough "live" examples of different ways to express the idea, so we could audit and have a conversation about what we think is most readable.

candida-d (approved):

PR #311, opened 2022-10-04
Wrap up margin examples + suggested edits

witgaw:


docs/concepts/trading-on-vega/positions-margin.md
```diff
@@ -92,7 +92,9 @@ The initial margin level being higher than the *margin search level* (which itse
 ### Margin level: Searching for collateral
 If the balance available in a trader's margin account is less than their position's *margin search level*, but is still above the maintenance level -- the network will try to allocate more money (up to the current initial margin level) from a trader's general account to be used for margin.
 
-If the margin account can be topped up, then the position stays open. If a market's swing is bigger than a user's margin is can cover, then money is pulled from the collateral to cover the requirement. In most cases, the allocated margin should cover market swings.
+If the margin account can be topped up, then the position stays open. If that's not possible as there are no sufficient funds in trader's collateral account and the market keeps moving against the trader so that the margin account balance drops below the current maintenance  level a closeout procedure is initiated. The first step is to cancel all of trader's open orders and revaluate the margin requirements. If the margin account balance is now higher than the maintenance margin level trader's position remains open, otherwise the protocol will attempt a liquidation.
```
candida-d, line 95:
```suggestion
If the margin account can be topped up, then the position stays open. If that's not possible as there are no sufficient funds in the trader's collateral account, and the market keeps moving against the trader so that the margin account balance drops below the current maintenance level, a closeout procedure is initiated. The first step is to cancel all of the trader's open orders and reevaluate the margin requirements. If the margin account balance is then higher than the maintenance margin level, the trader's position remains open, otherwise the protocol will attempt a liquidation.
```

candida-d (approved):

PR #370, opened 2022-11-04
Fix DataNodes table matching for some hosts & add refresh time

### LINKED ISSUE #352: DataNodes component does not tick a node where the GRPC hostname differs from REST
If a node has different hostnames for different endpoints (e.g. rest.testnet.vega.xyz vs grpc.testnet.vega.xyz), the GRPC endpoint will not get a ✅ despite being an 'available' node at build time.

# Tasks
- [ ] Improve hostname matching in `DataNode`

edd:
Fixes mistmatches between nodes with different subdomains for some APIs.
Also adds a refresh time, so it's clear when the connection star was last checked

- Fix DataNode host matching
- Add refresh time json
- Add refresh time below table

Screenshot 2022-11-04 at 17 42 35

Closes #352


candida-d (approved):
Great solution to highlight how fresh the data is.

PR #385, opened 2022-11-10
Concepts: data node

### LINKED ISSUE #372: Add network limits parameters & limitations
There will be some options for setting limits to certain transactions: pegged orders, tendermint transactions within a block, LP commitment size & order shapes

Dependent on parameters being created and spec being rewritten: https://github.com/vegaprotocol/specs/issues/1379

### LINKED ISSUE #373: Concepts: data nodes
What is a data node, what it provides - describe decentralised history & link to docs setup guide and readme: https://github.com/vegaprotocol/vega/blob/develop/datanode/dehistory/README.md

### LINKED ISSUE #374: Concepts: Enhance limited network lifetime
What, when, why (tendermint upgrades)

### LINKED ISSUE #388: Diagram: Data node
Turn the image below into a diagram that can be used in the docs for a high-level overview of the data node architecture. 

IMG_1984

candida-d:
Describes what data nodes are, details spam protection measures and adds more about network limits

Closes #372 
Closes #373 
Closes #374
Closes #388


docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,128 @@
...
+
+The values of all spam protection network parameters can be changed through a governance vote. If a parameter change passes governance, it takes effect in the epoch after it passes.
+
+<!-- link to governance -->
+
+### Proof of work 
+Although Vega is a proof-of-stake network, there is a client-side proof-of-work mechanism to prevent transaction spam from public keys trying to flood the network, or submitting a number of transactions that could slow down the network for all participants. The proof-of-work is calculated by the Vega wallet. It does not incur gas fees and does not have any effect on a transaction's priority.
+
+Every transaction must include a proof-of-work calculation derived from a recent block, proving the transaction was created recently. If a public key submits transactions above the maximum set by the network parameter (<NetworkParameter frontMatter={frontMatter} param="spam.pow.numberOfTxPerBlock" hideValue={true} />), the difficulty of the PoW increases and the proof will take longer to calculate, i.e., transactions will take longer to be generated but can still be included in a block.
```
ze97286, line 42: >If a public key submits transactions above the maximum set by the network parameter

This needs to be clearer that this refers to the number of transactions whose proof of work is based on a given block. 

>the difficulty of the PoW increases

This is only true if the network parameter for increaseDifficulty is not 0.

docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,128 @@
...
+
+It's possible to submit a large volume of transactions at one time, as long as the proof-of-work is calculated by the wallet adequately.
 
-There are several reasons for this decision, including: 
+A transaction with a missing or incorrect proof is rejected, as is any transaction that uses an already-used PoW calculation. A public key that sends transactions with faulty proof-of-work calculations is banned for 30 seconds or 1/48th of an epoch, whichever is greater.
```
ze97286, line 46: missing or incorrect proof is rejected but is rejected before a transaction is added to a block so it can't cause a party to be banned.

docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,128 @@
...
 
-### Network checkpoints 
+:::note Go deeper
+[Spam protection PoW ↗](https://github.com/vegaprotocol/specs/blob/master/protocol/0072-SPPW-spam-protection-PoW.md): Read the spec for implementation details and how this mechanism works in particular circumstances.
+:::
+
+### Spam policy enforcement
+Messages that don't follow the anti-spam rules are rejected, either pre- or post-block.
+
+Pre-block rejection: A transaction is rejected before it enters the validators' mempool. The case for a rejection can only be determined based on transactions that were in the previous block, and not on any other transactions received by a validator but not yet put into a block. Once a block is scheduled, all transactions in the mempool are verified. If a transaction fails the applicable spam rules set by the [parameters](#spam-protection-parameters) below, it's not processed and is removed.
```
ze97286, line 57: >Once a block is scheduled, all transactions in the mempool are verified

that's not true (or accurate). Transactions are pre-block checked for spam **before being added** to the mempool. The remaining transactions that were not selected for a block are rechecked after the block execution.

docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,128 @@
...
+
+Post-block rejection: A transaction makes it into the block, but is rejected before it's passed to the application layer. This allows for more fine-grained policies because more information is available, but the offending transaction will take up space in the block. If a transaction fails the spam rules set by the [parameters](#spam-protection-parameters) below, it is rejected and may trigger a spam proof of work threshold increase.
```
ze97286, line 59: >it is rejected and may trigger a spam proof of work threshold increase

the first part is correct, the second is not. It may cause the party to be banned and it may increase the min token requirement for submitting transactions but it may not do anything to pow threshold.

ze97286 (commented):
what about gas? is this explained somewhere else?

docs/concepts/vega-chain/network.md
candida-d: @ze97286 Min token requirement is just increased for the individual pubkey, right?

candida-d, line 59:
```suggestion
Post-block rejection: A transaction makes it into the block, but is rejected before it's passed to the application layer. This allows for more fine-grained policies because more information is available, but the offending transaction will take up space in the block. If a transaction fails the spam rules set by the [parameters](#spam-protection-parameters) below, it is rejected. The party may be banned, and the minimum token requirement for submitting transactions may be increased.
```

docs/concepts/vega-chain/network.md
candida-d: Thanks - suggested changes below :)

docs/concepts/vega-chain/network.md
candida-d: Ok, I think I see where I've misinterpreted, but I want to check, is this part of the spec below true, @ze97286?: 

> Once a block is scheduled, all validators also test all transactions in their mempool if they are still passing the test, and remove them otherwise.

pscott31 (commented):
The data node concepts stuff is really good :)

docs/concepts/vega-chain/data-nodes.md
```diff
@@ -0,0 +1,139 @@
...
+
+History segments produced on every node for a given block span are identical, such that the IPFS content IDs of the segment are the same across all nodes. This means there's no need to upload data between nodes, as each node produces history segments, and thus can be a source of history segments. 
+
+When a new node joins the network and requests a history segment for a given IPFS content ID, the provision of data to the node is shared across all other nodes in the network. The data node software also supports pulling all available historical data asynchronously once the newly joined data node has loaded the latest history segment. 
+
+A decentralised history segment is created every snapshot interval. It gets the interval's worth of data from the PostgreSQL database to create the history segment, without blocking event processing. The data node then publishes it to the network's IPFS swarm, thus making it available to all other current and future data nodes (per network).
+
+:::note Go deeper
+[Decentralised history readme](https://github.com/vegaprotocol/vega/blob/develop/datanode/dehistory/README.md): How to use decentralised history
+:::
```
candida-d, line 139:
```suggestion
## Network history
The daily addition of gigabites of core event data to the data node make it infeasible for a new data node to replay all blocks from the first block. Instead, new nodes can use a feature called network history to recreate the history quickly and get into a state to be able to consume new blocks. 

History segments produced on every node for a given block span are identical, such that the IPFS content IDs of the segment are the same across all nodes. This means there's no need to upload data between nodes, as each node produces history segments, and thus can be a source of history segments. 

When a new node joins the network and requests a history segment for a given IPFS content ID, the provision of data to the node is shared across all other nodes in the network. The data node software also supports pulling all available historical data asynchronously once the newly joined data node has loaded the latest history segment. 

A network history segment is created every snapshot interval. It gets the interval's worth of data from the PostgreSQL database to create the history segment, without blocking event processing. The data node then publishes it to the network's IPFS swarm, thus making it available to all other current and future data nodes (per network).

:::note Go deeper
[Network history readme](https://github.com/vegaprotocol/vega/blob/develop/datanode/networkhistory/README.md): How to use network history
:::
```

docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,178 @@
...
 
 Those checkpoints happen at defined intervals, and on every deposit and withdrawal request. 
 
-1. Each validator node calculates the hash of the checkpoint file and then sends this through consensus to ensure all the nodes in the network agree on the state. 
-2. If/when a network is taken down, the checkpoint file's hash is added to the genesis block. 
+### How a checkpoint is created and used
+1. Each validator node calculates the hash of the checkpoint file and then sends this through consensus to ensure all the nodes in the network agree on the state. A checkpoint file is then automatically produced and saved with a hash. 
+2. When a network is taken down, the checkpoint file's hash is added to the genesis block. A node operator updates the genesis file in the networks repo with the network's new start date, the new network ID, and the hash of the checkpoint file to be used in the network restoration. Once the file is updated, the node operators manually approve the change, and then one validator submits the restore transaction with the relevant checkpoint file.
```
ze97286, line 138: this is no longer true. There is no checkpoint restore command. The checkpoint data is embedded in the genesis file.

docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,178 @@
...
-4. All other validators verify the checkpoint against their own, restore the state and then allow other transactions on the network. 
+4. All other validators verify the checkpoint against their own. If the hash does not match, the transaction will have no effect. If the genesis file still has a previous state hash, all transactions will be rejected until the restore transaction arrives and is processed. If the hash matches, the network will be restored with the state and then allow other transactions on the network.
+
+### Validator scores in a restart
+The protocol needs a way to allow validators to continue initiating a restart, even before all information has been restored. 
+
+Each checkpoint includes the node IDs of all consensus, standby, and candidate validator nodes and their scores. 
+
+When initiating the restart all the nodes participating will temporarily be given the same Tendermint weight, which is used until the checkpoint has been fully processed. At that point, the weights are recalculated based on their state at the time the checkpoint was taken, and validators' scores are updated.
```
ze97286, line 147: again, no longer true.

docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,178 @@
...
+
+**Post-block rejection:** When a transaction makes it into the block, it is still checked before the block is finalised, and can then be rejected before it's passed to the application layer. A transaction may be rejected post-block and passed the initial validation if, for example, it breaches the maximum allowed rule, but nodes aren't yet aware of how many of that party's transactions will eventually be included in a block. That can happen because a validator node doesn't know what's in the other validators' mempools until transactions reach them through gossip. When the nodes evaluate a block during its execution, all mempool and transaction information is available, so any transactions that breach those spam rules are rejected and removed from the block before they can be executed.
```
ze97286, line 71: The first half of the sentence is inaccurate - essentially mempool are not part of this.

> When the nodes evaluate a block during its execution, all mempool and transaction information is available, so any transactions that breach those spam rules are rejected and removed from the block before they can be executed.

When blocks are executed, we have all the information from before the block + all transactions already seen within the block _so any transactions that breach those spam rules are rejected and removed from the block before they can be executed._

docs/concepts/vega-chain/network.md
```diff
@@ -1,27 +1,178 @@
...
+These limits are not totally dissimilar to the spam controls that are also implemented, which aim to mitigate the possibility of malicious actors deliberately flooding the system with junk requests. 
+
+### Maximum pegged orders and liquidity commitment shapes
+Currently, two transaction types can be limited:
+* Total number of pegged orders on all active markets: controlled by network parameter <NetworkParameter frontMatter={frontMatter} param="limits.markets.maxPeggedOrders" />
+* Number of offsets in each liquidity commitment order shape (per side), controlled by <NetworkParameter frontMatter={frontMatter} param="market.liquidityProvision.shapes.maxSize" />
+
+If either parameter's value is decreased (through a governance proposal and vote), then the change does not affect existing orders on the market, but only new orders/liquidity commitments placed after the change is enacted. 
+
+## Spam protection
```
ze97286, line 32: with all this talk about spam protection is it worth mentioning the spam statistics API? are they mentioned elsewhere? is it mentioned anywhere what guarantees does the wallet give you with regard to spam?

docs/concepts/vega-chain/data-nodes.md
```diff
@@ -0,0 +1,139 @@
...
+
+Alternatively, you can run your own data node and use [decentralised history](#decentralised-history) to pull existing data into your new database, or start fresh.
+
+:::caution Reliability
+PostgreSQL database schemas are liable to change without warning and without deprecation.
+:::
+
+<!-- Link to the data node postgres information.-->
+
+## Data retention
```
candida-d, line 89: @guoguojin Does anything in this section need to change based on https://github.com/vegaprotocol/vega/pull/7476?

guoguojin: Yes, you're missing a few data types:

trades_candle_1_day = 1 year
markets = 1 year
deposits = 1 year
withdrawals = 1 year
blocks = equal to the longest retention duration across all tables

The default retention period for lite = 1 day

Retention policy for any entity can be overridden by creating an entry in the config file under the SQLStore section:

```yaml
...
[SQLStore]
  [[SQLStore.RetentionPolicies]]
    HypertableOrCaggName = "balances"
    DataRetentionPeriod = "7 days"

  [[SQLStore.RetentionPolicies]]
    HypertableOrCaggName = "checkpoints"
    DataRetentionPeriod = "7 days"

...
```

guoguojin: Also, the retention policy for each entity is logged out when the node starts so they should be able to see what it is easily from there.

If they want to change the retention policy, a restart of the data-node is required.

docs/concepts/vega-chain/network.md
candida-d: as discussed off gh but here for posterity - I've added a tip box about the spam stats API, and will cover the wallet enhancements once they're merged

docs/concepts/vega-chain/data-nodes.md
```diff
@@ -0,0 +1,160 @@
...
+
+Some data can be saved in detail, and in a less-detailed sampling. Take balances, for example: In addition to saving every balance change, the `conflated balances` keep a sample of a balance from once every hour. This allows the data node to have data for a long period of time, but taking up less space than saving every balance change over a year.
+
+The data node code's configuration includes a set of default retention time frames. Those time frames have been defined with an individual trader that wants to run their own data node, for their own use, in mind. A public data node that's intended to support querying archival information will need much longer data retention time frames.
+
+Whenever a data node starts up, the existing retention policy for each data type is output in the logs.
+
+### Data retention modes
+When starting a data node, you can choose one of three retention modes, depending on what you want the data node to do.
```
guoguojin, line 101: This should be:

When initializing a data node.

When starting a data node, the chosen retention model will be applied. After initialization, it's possible to change the retention mode, for example from standard to archive, or lite by changing the retention period setting in the configuration file.

ze97286 (approved):

docs/concepts/vega-chain/network.md
```diff
@@ -130,10 +125,18 @@ See a full list of the network parameters used for PoW in the [spam protection p
 | spam.pow.increaseDifficulty           | If a wallet exceeds the max transactions per block, the difficulty of the PoW increases by this factor | <NetworkParameter frontMatter={frontMatter} param="spam.pow.increaseDifficulty" hideName={true} />
 | spam.pow.numberOfPastBlocks         	| To compute the wallet transaction quota, transactions can be assigned to past blocks; this parameter defines how far back that goes | <NetworkParameter frontMatter={frontMatter} param="spam.pow.numberOfPastBlocks" hideName={true} />
 
+
+## Loading network state: Snapshots
+To allow a Vega node to be restarted without the need to replay the whole blockchain, a Vega node can load an existing snapshot created by a different node, which will populate all the network state. The node can then resume listening to blocks after the snapshot, until it gets to the live block height, at which point it will be able to contribute to the chain.
```
davidsiska-vega, line 130: I'd say "node" state instead of "network" state.

davidsiska-vega: "resume listening to blocks after the snapshot, until ..."

I'd say 

"downloads and replays blocks from the height corresponding to snapshot, until ..."

docs/concepts/vega-chain/network.md
```diff
@@ -130,10 +125,18 @@ See a full list of the network parameters used for PoW in the [spam protection p
...
+
+:::note Try it out
+Node operators can refer to [how to use snapshots](../../node-operators/how-to/use-snapshots.md) for a step-by-step guide on loading network state with snapshots. 
+:::
+
 ## Network restarts: Checkpoints
```
davidsiska-vega, line 136: "Checkpoints" I like to say LNL Checkpoints to make it even more different from Snapshots to disambiguate.

davidsiska-vega (approved):

PR #422, opened 2022-12-10
Fix REST v2 documentation in 0.64.0

edd:
# Problems
Screenshot 2022-12-10 at 09 14 55
- v2 is not the correct name
- v1 should have been removed

Screenshot 2022-12-10 at 09 15 22
- rest pages crash

# Fixes
| **Problem** | **Fix** |
|-------------|---------|
|  `docusaurus-plugin-openapi-doc`I *think* was crashing because the version had moved on, but docusaurus had not.  | pin version of `docusaurus-plugin-openapi-doc` (i.e. remove `^`)        |
|  When `trading_data_v1.swagger.json` stopped existing, because we generate the configuration for `docusaurus-plugin-openapi-doc` based on the files, the plugin no longer removed it when `clean-api-doc` was called. | #425   |
|  `sidebars.js` still imported `trading_data_v1/sidebar.js` which still existed (see above) | Removed after removing `docs/api/rest/trading_data_v1` started failing the build |
|  `api/rest/overview.md` and one other page still referenced v1 apis as soon-to-be-deprecated | Removed after removing `docs/api/rest/trading_data_v1` started failing the build |
| Core and core state services had wrong urls | #424  |

# Follow up
- This isn't the first time we've had `This page has crashed` pages served while updating API versions. The error isn't in the built.html, it happens when the client-side JS renders. So it can't be detected by `ack`ing over the built files. But some sort of automated check could be written... #426 
- Need to detect if APIs are removed and flag it somehow
- Update build script not to inject data node URLs in to core/core state swaggers/openapis


candida-d (approved):
seems better at least

PR #503, opened 2023-02-24
chore: minor changes to migration guide

fkondej:


docs/node-operators/migration-guides/upgrade-node.md
```diff
@@ -186,7 +188,7 @@ While Visor is optional, it is strongly recommended that you install and use Vis
 If you will NOT use Visor, skip to [step 12](#12-create-vega-and-data-node-systemd-services).
 
 ### 8. Initiate Visor
-It's strongly recommended that you set up Visor for automatic protocol upgrades. Visor enables you to upgrade to future versions without needing to coordinate with all validators. If you have Visor enabled, a validator can propose a block for an upgrade, your node can agree to that proposal, and the upgrade will happen at the predetermined block height without intervention.
+It's strongly recommended that you set up Visor for automatic protocol upgrades, i.e. upgrade your node to future version. With visor the upgrade will happen at the predetermined block height without manual intervention.
```
candida-d, line 191:
```suggestion
It's strongly recommended that you set up Visor for automatic protocol upgrades, i.e. to upgrade your node to a newer version. With Visor, an upgrade happens at a predetermined block height without manual intervention.
```

docs/node-operators/migration-guides/upgrade-node.md
```diff
@@ -253,12 +255,12 @@ name = "genesis"
 
 [vega]
   [vega.binary]
-    path = "<VEGA-BIN>"
+    path = "vega"
```
candida-d, line 258: @fkondej - there's a variable at the top of the page for the vega binaries path. With that in mind, do you still think it should be changed?

> `<VEGA-BIN>`: the path to the Vega core binary, e.g., `/home/vega/bin/vega`

fkondej: It is up for discussion, but I don't think we should use `<VEGA-BIN>`.
We need to consider there will be more than one `vega` binary. E.g. for the `Protocol Upgrade`, there will be at least two: one before the upgrade and one after the upgrade.

Vegavisor proposes an approach that there should be a subdirectory for every vega version under `<VEGAVISOR-HOME>`. Each subdirectory is called the same as the version, where `genesis` is a special case. And each subdirectory contains the `vega` binary. And `run-config.toml`, i.e. the file for this example/snippet. When we put `vega` as value for the `path` parameter, we point to a binary called `vega` inside the same directory as `run-config.toml`.
On the other hand, `<VEGA-BIN>` starts with `/`, so it is an absolute path. Technically this can be correct, but does not prepare you there will be more than one `vega` binary.

docs/node-operators/migration-guides/upgrade-node.md
```diff
@@ -90,7 +90,10 @@ cp -r <VEGA-NETWORK-HOME>/config <BACKUP-FOLDER>/v0.53.0/vega-config
 cp -r <TENDERMINT-HOME>/config <BACKUP-FOLDER>/v0.53.0/tendermint-config
 
 # copy wallets
-cp -r <VEGA-NETWORK-HOME>/data/node/wallets <BACKUP-FOLDER>/v0.53.0/core-wallets
+cp -r <VEGA-NETWORK-HOME>/data/node/wallets <BACKUP-FOLDER>/v0.53.0/wallets
+cp <TENDERMINT-HOME>/node_key.json <BACKUP-FOLDER>/v0.53.0/wallets
+cp <TENDERMINT-HOME>/priv_validator_key.json <BACKUP-FOLDER>/v0.53.0/wallets
+cp <VEGA-NETWORK-HOME>/nodewallet-passphrase.txt <BACKUP-FOLDER>/v0.53.0/wallets  # filename and locaiton might differ, depending on your setup
```
candida-d:
```suggestion
cp <VEGA-NETWORK-HOME>/nodewallet-passphrase.txt <BACKUP-FOLDER>/v0.53.0/wallets  # filename and location might differ, depending on your setup
```

candida-d (approved):

daniel1302 (approved):

PR #530, opened 2023-03-13
Describe slippage market parameters

candida-d:


docs/concepts/trading-on-vega/positions-margin.md
```diff
@@ -75,6 +75,17 @@ The amount a trader will have held aside as maintenance margin is derived from t
...
+Maintenance margin is calculated as: 
+
+```
+maintenance margin = price x (linear slippage factor x |position| + quadratic slippage factor x position^2) + price x |position| x size x risk factor
+```
+
+Slippage factors are market parameters that specify by how much the liquidity component of the margin calculation is dependent on the position size in a low-volume market scenario.
+
+If there is enough volume on the book, the slippage comes directly from the book and the liquidity component is not used. Margin slippage in a low-volume scenario is calculated as `slippageFromFactors = linear x position  + quadratic x position^2) x price`. If there is a lot of liquidity on the book, the protocol calculates the closeout amount and provides the lower amount, i.e., the liquidity part of the margin `min(slippageFromFactors, slippageFromBook)`. Increasing the linear or quadratic slippage factor increases the liquidity part of the margin calculation, but only if there is little volume on the book; if there is enough volume on the book the slippage comes directly from the book.
+
```
davidsiska-vega, line 88:
```suggestion

The `risk factor` will be different for short and long positions; the risk model provides `risk factor long` for when `position > 0` and `risk factor short` for when `position < 0`. 

Note that your party's limit orders are included as well and the maintenance margin is calculated for the riskiest long or short combination of orders and position. 
```

davidsiska-vega (changes requested):
I've left a suggestion but otherwise LGTM.

PR #533, opened 2023-03-15
update governance site URL for fairground

candida-d:


edd:
This URL came from https://github.com/vegaprotocol/networks-internal/blob/main/fairground/vegawallet-fairground.toml fyi. This update works, but will get overwritten next time `./scripts/build.sh` is run (i.e. next tag). I've poked someone to update the `.toml`.

edd (approved):

PR #537, opened 2023-03-16
Add page for websocket

### LINKED ISSUE #536: Describe how to stream data with WS
- [ ] Add more detail to individual reference endpoints for in-place guidance
 - [ ] Create new page with walkthrough guidance for how to use the endpoints
 - [ ] Link to exhaustive list of each endpoint, as they're not easily discoverablein the trading-data list
 - [ ] (Maybe) filter them out of trading-data list and display them in their own section

candida-d:
closes #536 

Guide people on how to use the websocket APIs available

### DIFF
--- a/docs/api/websocket.md
+++ b/docs/api/websocket.md
@@ -2,7 +2,7 @@
 title: WebSocket streams
 sidebar_label: WebSocket streams
 hide_title: false
-description: Learn how to use the available websocket APIs.
+description: Learn how to use the available WebSocket APIs.
 vega_network: TESTNET
 ---
 import Tabs from '@theme/Tabs';
@@ -85,7 +85,7 @@ client.onmessage = console.dir;
 </TabItem>
 </Tabs>
 
-The above examples show how to use Websockets to stream all ledger movements that occur on the Vega Fairground network. An example payload is shown below:
+The above examples show how to use WebSockets to stream all ledger movements that occur on the Vega Fairground network. An example payload is shown below:
 
 ```json
 {
@@ -127,7 +127,7 @@ The above examples show how to use Websockets to stream all ledger movements tha
 
 All enum values are sent as their integer values and not their string representation. This is to reduce the amount of data in each packet and for speedy performance.
 
-## Snapshot data
+## Snapshots of data
 
 Some of the WebSocket endpoints will send a snapshot of the current state of data when a connection is first made. This allows for an application to build an initial state, creating context for subsequent updates. The snapshot data will be sent in batches after which subsequent messages will only be updates to the snapshot state.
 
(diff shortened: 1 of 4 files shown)


wwestgarth:
A few things to note:
- I removed `## Available websocket APIs` because 1) I didn't know how to do links!! and 2) they might all change anyway when we reshuffle the protos/move the subscription endpoints into their own section
- The new page doesn't appear in the sidebar anywhere, I wasn't sure where it should live
- We have a problem with the enums being ints instread of strings: I don't think there is anywhere in the autogenerate docs that would tell someone what each int means!

docs/api/websocket.md
```diff
@@ -0,0 +1,211 @@
...
+
+Vega's API contains websocket endpoints that offer real-time updates to changes in state of the Vega network, allowing subscriptions to events such as per-market trades or changes to a party's position.
+
+As Vega is a blockchain time moves in discrete blocks and so updates will appear as blocks are executed.
+
+## Authentication and Ratelimiting
+
+API tokens are not required to access the API as they are all public. TLS is supported on all websocket endpoints but note that whether it is enabled on a particular data node is a choice made by the node operator.
+
+Websocket connections are ratelimited by a maximum allowed number of subscriptions per IP address. The default maximum is set to 250 connections but note that this value may differ between data node operators.
```
edd, line 17:
```suggestion
Websocket connections are rate limited by a maximum allowed number of subscriptions per IP address. The default maximum is set to 250 connections but note that this value may differ between data node operators.
```

docs/api/websocket.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+The last batch of snapshot data will have `lastPage` set to `true` after which the stream will switch to sending updates to the orderbook
+
+
+```json
+{
+    "result": {
+        "updates": {
+            "orders": [],
```
edd, line 145:
```suggestion
            "orders": []
```

As above

edd (approved):
I haven't run the code samples. Two validation things in JSON, one pedantic thing that @candida-d can agree/disagree with

docs/api/websocket.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+</TabItem>
+
+<TabItem value="js" label="Node">
+
+```js
+var WebSocketClient = require('websocket').client;
+
+url = "wss://api.n07.testnet.vega.xyz/api/v2/stream/ledger/movements"
```
edd, line 64:
```suggestion
var url = "wss://api.n07.testnet.vega.xyz/api/v2/stream/ledger/movements"
```

edd (changes requested):
JS is broken. I'll fix it.

edd (approved):

PR #542, opened 2023-03-17
docs: add sketch for restarting vega from remote snapshot

daniel1302:
This is only draft, that needs to be improved. I have committed it as one of the validator asked for instructions and I have prepared it for them


daniel1302 (commented):
We will not merge this PR. We have new one with better steps and more infos for mainnet:

https://github.com/vegaprotocol/documentation/pull/672

PR #543, opened 2023-03-20
chore: add information about configuring chunk intervals  [v.0.70]

guoguojin:
Adding information regarding overriding chunk intervals for Timescale tables in data node.


candida-d (approved):
Ready to go once 0.70 is on testnet

PR #612, opened 2023-05-10
chore: update release summary notes with more detailed information

gordsport:


gordsport (commented):
@candida-d 
Ive added links to the vega docs pages in some instances as suggested. This is in place of the specs links - I can reinstate these if we wish

PR #620, opened 2023-05-12
fix: example timestamps should be in seconds, not milliseconds, since epoch

pscott31:
In the generated examples of proposals, the example code given was passing in timestamps as number of milliseconds since the epoch, rather than number of seconds which is what is actually required.


edd (changes requested):
The enactment/settlement dates shown in the metadata need to not be in the 1970s. These are used in the console, I believe!

Also: PR description please

edd (approved):

PR #629, opened 2023-05-23
Initial documentation of browser wallet

### LINKED ISSUE #528: Document the browser extension integration in the wallet API section

emilbayes:
Closes #528


docs/api/vega-wallet/reference/browser-extension/javascript-api.md
```diff
@@ -0,0 +1,260 @@
...
+a high-level Javascript API, injected into web pages.
+
+### Detecting the extension
+
+The Vega browser wallet injects methods on the global `vega` object,
+hence the presence of the extension can be detected by checking that
+this object is defined:
+
+```js
+if (globalThis.vega != null) {
```
dexturr, line 24: The eth docs seem to use window, i guess we should be using this as lowest common denominator if possible.

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
```diff
@@ -0,0 +1,260 @@
...
+different entry points into the network or different networks entirely 
+(eg. `mainnet`, `fairground`, etc.)
+
+A dApp can detect the network in use by checking the `chainID` in use:
+
+```js
+const { chainID } = await vega.getChainId()
+```
+
+Note that `chainID` is not an indefinate constant but may change with
```
dexturr, line 67: nice

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
```diff
@@ -0,0 +1,260 @@
...
+
+- [Type Definition](https://github.com/vegaprotocol/js-protos/blob/main/dist/vega/commands/v1/DelegateSubmission.d.ts)
+- [Protobuf Definition](../../../grpc/vega/commands/v1/commands.proto.mdx#delegatesubmission)
+
+```js
+{
+  // string; Stake to Marvin on Fairground
+  nodeId: 'bb1822715aa86ce0e205aa4c78e9b71cdeaec94596ce72d366f0d50589eb1bf5',
+  // string; 1 $VEGA
+  amount: (1n * 18n).toString()
```
dexturr, line 171: @candida-d do we have any existing docs about how numbers are strings in vega to prevent flop errors? Can cross link to that here?

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
```diff
@@ -0,0 +1,260 @@
...
+
+Transactions in Vega are protocol buffers. The browser wallet provides a JSON
+interface which translates to protobuf fields. As a general rule fields are
+camel-cased, falsy values are not encoded (they map to protobuf default values)
+and optional values can be set to `null` to signify explicit absence.
+
+Below are sample transactions and links to their corresponding `.d.ts` files and
+protobuf comments.
+
+#### Transfer
```
dexturr, line 126: @candida-d if we have other documenation for commands should we just link it here rather than duplicate? Guess it depends if the transactions are identical shapes 👍

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
dalebennett1992: Sorry to be that guy but should it be 'indefinite'?

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
emilbayes: Yeah good call. I used `globalThis` as that's `window` but also works in WebWorkers and ServiceWorkers, but I don't know if extensions are exposed in such contexts

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
emilbayes: I think they have identical shapes from what I can tell apart from places where protobuf expects (u)int64, in which case web wallet accepts bigints while desktop just takes normal numbers from what I can tell. My intention was for this to be a quick overview, (probably should have been a "getting started" rather than API docs), so I thought it would be good with some tangible examples rather than straight up docs. However I agree, I will cross link to the other docs

docs/api/vega-wallet/reference/browser-extension/get-started.md
```diff
@@ -0,0 +1,272 @@
...
+```
+
+The first time a dApp tries to connect to a wallet, the user will be
+prompted to approve the connection and select which keys they will
+allow the dApp to see. Only appoved keys will be available to the dApp.
+
+Subsequent connect attempts will resolve automatically, unless the user
+has removed the connection in the meantime.
+
+Once a dApp is finished talking to the wallet it can call
```
candida-d, line 51: Something to consider - what about adding guidance or suggestions on when it would be a good time to automatically 'disconnect wallet'?

emilbayes: Yeah, I don't actually know if you would ever do that automatically or just offer it as an option to the user. I at least can't think of a reason other than to be "nice" if the dApp does some kind of one-off transaction

docs/api/vega-wallet/reference/browser-extension/get-started.md
```diff
@@ -0,0 +1,268 @@
...
+})
+```
+
+`sendingMode` decides when the Vega node returns a response to the transaction
+submission. `TYPE_ASYNC` will run validation checks by the network in cases 
+where this is possible, but will return fast. `TYPE_SYNC` is the perferred mode.
+This mode will return once the transaction has passed validation checks and 
+entered the mempool.  `MODE_COMMIT` is only available for debug purposes and 
+will require your own node with this mode enabled. This will cause the 
+submission to wait until the transaction has been mined.
```
candida-d, line 109:
```suggestion
`sendingMode` determines when the Vega node returns a response to the transaction
submission. 

`TYPE_ASYNC` will run validation checks by the network in cases  where this is possible, but will return fast. 

`TYPE_SYNC` is the perferred mode. This mode will return once the transaction has passed validation checks and entered the mempool.  

`MODE_COMMIT` is only available for debug purposes and will require your own node with this mode enabled. This will cause the submission to wait until the transaction has been mined.
```

candida-d: I'd move sync up to be first since it's the preferred one

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
```diff
@@ -0,0 +1,70 @@
+---
+title: Build with the browser extension
```
candida-d, line 2:
```suggestion
title: Javascript API
```

candida-d:
@emilbayes As you work on this, can you also check the pages outside of the reference folder to make sure they refer to the browser wallet/docs in the relevant places? Right now they're all geared towards using the JSON-RPC API.

docs/api/vega-wallet/reference/browser-extension/javascript-api.md
candida-d: We do! However they were published after this branch was created, so they won't be available without a rebase/merge from main.

emilbayes:
@candida-d rebased on latest `main` and fixed a few comments. I will continue documenting `sendTransaction(...)`

PR #656, opened 2023-06-02
grammatical fix

CommodumIan:
I stumbled across this as I was reading the doco.


candida-d:
Thanks very much @CommodumIan! 

If you have a moment, could you make the same change on this page, too? https://github.com/CommodumIan/documentation/blob/main/versioned_docs/version-v0.71/concepts/trading-on-vega/positions-margin.md

Same content, but that folder is where the mainnet docs are served from.

candida-d (approved):

PR #689, opened 2023-06-30
Concepts: successor markets

### LINKED ISSUE #631: Concepts: Successor markets
* Describe successor markets in concepts
* Cover any changes to governance & tutorials (as needed)

candida-d:
Closes #631


docs/concepts/liquidity/provision.md
```diff
@@ -32,7 +32,9 @@ Liquidity providers need to be able to support their liquidity commitment - thei
...
+Participants who want to commit liquidity to a market can enter their commitments as soon as a market proposal is submitted and accepted, even before the governance vote to create the market concludes, as well as at any time while the market is trading. Committing earlier in a market's lifecycle leads to a higher [equity-like share](rewards-penalties.md#liquidity-fees) in that market, assuming the trading volume on the market increases with time. 
+
+A liquidity provider's equity-like share is then carried over to the successor market of the market they've provided liquidity for, if a successor market comes to exist. The provider will then need to provide liquidity on the successor market while it's [pending](../trading-on-vega/market-lifecycle.md#market-status-pending)to keep that equity-like share active.
```
edd, line 37:
```suggestion
A liquidity provider's equity-like share is then carried over to the successor market of the market they've provided liquidity for, if a successor market comes to exist. The provider will then need to provide liquidity on the successor market while it's [pending](../trading-on-vega/market-lifecycle.md#market-status-pending) to keep that equity-like share active.
```

Adding a missing space

edd (commented):
A few questions, one nitpicky change - but the main body of everything is nice and clear

docs/concepts/governance.md
```diff
@@ -228,6 +235,23 @@ Another parameter is
...
+* Market ID of the parent market **must** be defined
+* Insurance pool percentage: A percentage of the parent market's insurance pool, up to 100%, can be earmarked for transfer to the successor market
+
+The following details will need to match between parent and successor:
+* Settlement asset
+* Quote: Determined from the data source
+
+For a successor market to be enacted, the parent market must be in one of the following states: proposed, pending, active, suspended or trading terminated. 
+
+The parent market can be settled or cancelled when the successor market reaches enactment time, as long as the time it's been settled/cancelled is equal to or less than the parent market's settlement time plus the `market.liquidity.successorLaunchWindowLength` - determined by a network parameter. This parameter specifies for how long after a market has settled, the liquidity provider's equity-like share data are retained and the insurance pool is left undistributed to allow a successor to be defined. If the successor is proposed after that time, then nothing is transferred over and the market opens as a standard market.
```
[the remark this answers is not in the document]
candida-d: @edd I will, once there is a netparam to link to, but as this feature isn't on the network yet, wouldn't it just fail to build properly and error?

docs/concepts/governance.md
```diff
@@ -228,6 +235,23 @@ Another parameter is
...
 <!--### Changing models [WIP]  -->
 
+### Propose a successor market
+A successor market is a market that will carry on after the original market, or parent, that it is based on has settled - though a parent and successor market can be active simultaneously. Proposing a new successor market that follows from an existing market allows liquidity providers to the option to keep their [equity-like share](./liquidity/rewards-penalties.md#how-liquidity-fees-are-split) on the new market, even when the original market expires. Creating an entirely new market with no parent doesn't offer the same benefit.
+
+Each market can have only one active successor.
+
+In terms of the proposal format, there are only two differences between a succesor market proposal and that for a regular market.  
+* Market ID of the parent market **must** be defined
+* Insurance pool percentage: A percentage of the parent market's insurance pool, up to 100%, can be earmarked for transfer to the successor market
```
candida-d, line 245:
```suggestion
* Parent market ID: Required to define the proposal as for a successor market
* Insurance pool percentage: A percentage of the parent market's insurance pool, up to 100%, can be earmarked for transfer to the successor market. A number between 0 and 1 (which represents the factor for the percentage) is required.
```

candida-d: @edd Is this better?

edd: Yes

docs/concepts/governance.md
```diff
@@ -228,6 +235,23 @@ Another parameter is
 
 <!--### Changing models [WIP]  -->
 
+### Propose a successor market
+A successor market is a market that will carry on after the original market, or parent, that it is based on has settled - though a parent and successor market can be active simultaneously. Proposing a new successor market that follows from an existing market allows liquidity providers to the option to keep their [equity-like share](./liquidity/rewards-penalties.md#how-liquidity-fees-are-split) on the new market, even when the original market expires. Creating an entirely new market with no parent doesn't offer the same benefit.
```
TomMcL, line 239:
```suggestion
A successor market is a market that will carry on after the original market, or parent, that it is based on has settled - though a parent and successor market can be active simultaneously. Proposing a new successor market that follows from an existing market offers liquidity providers the option to keep their [equity-like share](./liquidity/rewards-penalties.md#how-liquidity-fees-are-split) on the new market, even when the original market expires. Creating an entirely new market with no parent doesn't offer the same benefit.
```

docs/concepts/governance.md
```diff
@@ -228,6 +235,23 @@ Another parameter is
...
+
+Each market can have only one active successor.
+
+In terms of the proposal format, there are only two differences between a succesor market proposal and that for a regular market.  
+* Parent market ID: Required to define the proposal as for a successor market
+* Insurance pool percentage: Required percentage of the parent market's insurance pool, up to 100%, can be earmarked for transfer to the successor market. It is submitted as a number between and including 0 and 1, which represents the factor for the percentage.
+
+The following details will need to match between parent and successor:
+* Settlement asset
+* Quote: Determined from the data source
```
TomMcL, line 249: Quote?

docs/concepts/governance.md
```diff
@@ -41,7 +41,14 @@ The network will also calculate:
...
-    
+
+### Proposal outcome: Successor market
+For proposals adding a new successor market, the outcome of the proposal can change even after it's been approved. 
+
+If a parent market is still in its proposed state, its successor market cannot be enacted, even if it passes the vote.
+
+Two proposals that name the same parent can be submitted and pass a governance vote. Whichever market clears its [opening auction](./trading-on-vega/trading-modes.md#auction-type-opening) first gets the share of the insurance pool, and the liquidity providers' equity-like share is moved to that market. The second market will still open for normal trading, assuming it gets out of opening auction, but it will not be considered a successor market.
```
TomMcL, line 50: Second market will be cancelled I believe

docs/concepts/governance.md
candida-d: @TomMcL I'll take suggestions, as I don't think "quote asset" will necessarily always be true. Quote name?

docs/concepts/governance.md
candida-d: This is what the spec says, but I can check on the actual implementation if you think this is wrong. 

> if there are two proposals naming the same parent market then whichever one gets into the active state first (i.e. > passes governance vote and clears the opening auction) becomes the successor of the named parent; the other > proposal is kept but the parent market id field is cleared and when opening auction ends no virtual stake will get > carried over.

docs/concepts/governance.md
TomMcL: I don't think that has to be the same at all? A rolling 3m future would be settling on a very different oracle which doesn't inherently have a tie to the previous one

docs/concepts/governance.md
TomMcL: Seems we have some spec contention as there is an AC saying:


> With two successor markets in opening auction, that have the same parent market, and one additional market in the state "Proposed". Get one of the two markets to leave the opening auction (passage of time, LP commitment, crossing trade). The other market in auction and the proposed market should both be "Rejected" and all LP funds will be released ([0081-SUCM-014](https://github.com/vegaprotocol/documentation/pull/689#0081-SUCM-014))

I guess we need to decide which version they should be doing (I don't have a strong preference either way so perhaps lean towards whatever the code does @EVODelavega )

candida-d: Yep and there's another one saying: 

Two proposals that name the same parent can be submitted. Both can be approved by governance. The proposed market that clears the opening auction first gets the share of the insurance pool and the virtual stakes get carried over. Once the first market clears the opening auction the other market's parent market Id field is cleared. When it clears the opening auction it gets no insurance pool from the parent and no virtual stakes get carried over ([0081-SUCM-005](https://github.com/vegaprotocol/specs/blob/cosmicelevator/protocol/0081-SUCM-successor_markets.md#0081-SUCM-005)).

PR #699, opened 2023-07-07
chore: improve gRPC overview page

### LINKED ISSUE #623: Describe how to use gRPC docs
The docs could use a page that introduces gRPC, provides tips on how to use the proto docs directly (which provides a richer and more useful experience than the generated pages), and also explains to people how to refer to the REST docs, which provide the same information, if they need a docs page. 

Potentially, this could also include either hiding the gRPC generated pages from the sidebar, or getting rid of them completely if the docs can support the gRPC users we currently know of. 

Loosely speaking, the WebSocket page is a good example of what style of information to include. https://docs.vega.xyz/mainnet/api/websocket

wwestgarth:
closes #623 

I've gone a bit deeper into what gRPC is and dropped lines that suggest that theres no reason to look at our generated docs because the protos/REST docs will be enough...

Still TODO:
- [x] make sure the grpc client stub examples _actually work_
- [x] work out if we can do anything to stop repeating information about rate-limiting


docs/api/grpc/overview.md
```diff
@@ -1,16 +1,162 @@
...
+}
+```
+
+The above protobuf definitions can be used to generate client and server code in a given language. This can be done by using the protocol-compiler `protoc` directly and supplier a output language, or by using CLI tools from the buf-ecosystem, namely `buf generate`. As an example, YAML configuration for compiling Vega's protobuf definition into Golang using `buf genreate` can be found here [here](https://github.com/vegaprotocol/vega/blob/develop/buf.gen.yaml)
+
+
+Once generated they can imported and used from the generated language specific client stubs as follows:
+
+<Tabs>
+<TabItem value="py" label="Python">
```
wwestgarth, line 63: I've done python and golang examples. I don't know if we know what other languages might be useful given our knowledge of people who have already integrated?

candida-d: @wwestgarth Seems like this will do for now, and we can add more as we get requests/more data

edd (approved):

docs/api/grpc/overview.md
```diff
@@ -1,16 +1,152 @@
...
+## gRPC API and Protocol Buffers
+gRPC uses Protocol Buffers (protobufs) as its underlying data definition. Protobuf messages are language-neutral data definitions that are used to generate language-specific data structures. Code generation of a protobuf message also generates functions to serialise each message so that they can be sent through gRPC to each RPC service endpoint.
 
-## Data node API
+This means that Vega's gRPC API is fully defined by its protobuf definitions, and so the protobuf definitions themselves provide complete documentation.
+
+## Relationship to REST API
+The buf ecosystem contains plugins that also allow automatic generation of REST API endpoints from the protobuf definitions. The result is that Vega's gRPC and REST API match exactly in both structure and functionality.
+
+As REST is a more familiar way to interact with a product than gRPC, experiencing Vega first through the REST API and using the REST documentation may help ease an initial integration, without any loss of functionality. Migrating later to using gRPC is then trivial since the mapping between REST and gRPC is 1-to-1.
```
cdummett, line 21: One thing unclear to me (mostly because I don't know how the gRPC plugins work) is how the REST  endpoints are generated from the protobuf definitions - I don't think I would be able to make the connection between specific gRPC messages and REST endpoints without the REST docs.

Is this saying I should find it trivial to work-out what a REST endpoint is just from the proto definitions?

wwestgarth: I think the idea is that most people would go the other way around -- they'll play with the REST API first because its more familiar, and then work out how to go from the REST docs to gRPC. So in that situation they'll be looking at, say, the docs for this REST endpoint:
https://docs.vega.xyz/mainnet/api/rest/data-v2/trading-data-service-get-order
which has the heading "Get Order", wonder how to do it in gRPC and start tapping `.GetO` into their IDE which would autocomplete into `.GetOrder()` and they go "wow that was easy, I don't even _need_ to look at the gRPC docs"

If you were to want to go the other way and go from a gRPC call to the REST URL for it, the true answer is that the map is in this file which we could consider:
https://github.com/vegaprotocol/vega/blob/develop/protos/sources/data-node/grpc-rest-bindings.yml
but you could also just look through in the REST API docs to find it and the way they are grouped in the sidebar should help.

@candida-d do you think we should add a link to https://github.com/vegaprotocol/vega/blob/develop/protos/sources/data-node/grpc-rest-bindings.yml ? I'm worried it'll just make it seem more confusing.

docs/api/grpc/overview.md
```diff
@@ -1,16 +1,152 @@
...
+	case codes.OK:
+		fmt.Println("found order:", response.Order)
+		return
+	case codes.NotFound:
+		fmt.Println("order not found")
+		return
+	default:
+		fmt.Println("unable to get order from a data node:", err)
+	}
+}
```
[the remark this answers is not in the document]
wwestgarth: Yea fair enough, I've just going to remove these lines though because I don't think they are reachable:
```
case codes.OK:
    log.Fatal("found order:", response.Order)
```

docs/api/grpc/overview.md
```diff
@@ -1,16 +1,154 @@
...
+## gRPC API and Protocol Buffers
+gRPC uses Protocol Buffers (protobufs) as its underlying data definition. Protobuf messages are language-neutral data definitions that are used to generate language-specific data structures. Code generation of a protobuf message also generates functions to serialise each message so that they can be sent through gRPC to each RPC service endpoint.
 
-## Data node API
+This means that Vega's gRPC API is fully defined by its protobuf definitions, and so the protobuf definitions themselves provide complete documentation.
+
+## Relationship to REST API
+The buf ecosystem contains plugins that also allow automatic generation of REST API endpoints from the protobuf definitions. The result is that Vega's gRPC and REST API match exactly in both structure and functionality.
+
+As REST is a more familiar way to interact with a product than gRPC, experiencing Vega first through the REST API and using the REST documentation may help ease an initial integration, without any loss of functionality. Migrating later to using gRPC is then trivial since the mapping between REST and gRPC is 1-to-1.
```
candida-d, line 21:
```suggestion
If you are new to interacting with gRPC, using REST may be a more familiar way to interact with the APIs. Experiencing Vega first through the REST API and using the REST documentation may help ease an initial integration, without any loss of functionality. Migrating later to using gRPC is then trivial since the mapping between REST and gRPC is 1-to-1.
```

docs/api/grpc/overview.md
```diff
@@ -1,16 +1,206 @@
...
 
-## Data node API
+This means that Vega's gRPC API is fully defined by its protobuf definitions, and so the protobuf definitions themselves provide complete documentation.
+
+## Relationship to REST API
+The buf ecosystem contains plugins that also allow automatic generation of REST API endpoints from the protobuf definitions. The result is that Vega's gRPC and REST API match exactly in both structure and functionality.
+
+If you are new to interacting with gRPC, using REST may be a more familiar way to interact with the APIs. Experiencing Vega first through the REST API and using the REST documentation may help ease an initial integration, without any loss of functionality. Migrating later to using gRPC will then require minimal changes since the the input parameters and responses are contain the same data.
+
+As an example to show the similarities, below are two Python snippets of how to list transfers using both REST and gRPC:
```
candida-d, line 23:
```suggestion
As an example to show the similarities, below are two Python snippets for how to list transfers using both REST and gRPC:
```

docs/api/grpc/overview.md
wwestgarth: I've added two snippets showing how similar the using REST and gRPC are (and linked to the yml file that maps the url <-> gRPC function call), which is hopefully is more useful

PR #724, opened 2023-07-26
Add verified price guidance to tutorials

candida-d:


daunatv (approved):
This should surely be enough to help users creating a market which uses verified price as settlement.

PR #826, opened 2023-11-15
new layout with icons

candida-d:


edd:
Screenshot 2024-01-05 at 11 41 03

Homepage now looks a bit like this. i.e.:

- Black header section gone
- Feature List added to homepage, but with the same content as the concepts/basics page copy and pasted. See `src/pages/index.js` to change that content

PR #868, opened 2024-01-04
Hide submit button for core state endpoints

candida-d:
Closes #855

PR #871, opened 2024-01-05
fix: improve build script that inserts hide_send_buttons

edd:
improve hiding API send buttons 

The build script previously blindly inserted hide_send_button, which was a problem when it was duplicated. Now it
actually checks if the line is there, and only inserts it if it isn't. 

- Hide send button on mainnet core state APIs (as none are exposed)


edd:
Accidentally duplicates / replaces #868

edd:
This apparently disables the button in some places it shouldn't. The script change should be merged in to https://github.com/vegaprotocol/documentation/pull/868

PR #900, opened 2024-01-30
Liquidation updates

### LINKED ISSUE #812: Changes to liquidation & margin relationships
## New market proposal fields

Describe the new fields in a market proposal for liquidation changes:

- `disposal time step` -   "Specifies the interval, in seconds, at which point the network will try to unload its position."
 - `disposal fraction` -   "Specifies the fraction of its position the network will try to reduce its position by in a single disposal attempt."
 - `full disposal size` -   "Specifies the size of the position held by the network that it will try to dispose of in one attempt."
 - `max fraction [of book side within liquidity bounds] consumed` -   "Specifies the maximum size by which the network can reduce its position as a fraction of the volume on the book."

## Feature overview
The aim here is:

* Separate margin checks, mark-to-market and closeout trades into three separate streams that can run at arbitrary frequencies,
* Accommodate closeout trades during auctions,
*  Avoid allowing parties with huge position breaching margin level hanging around.
*  Do this without changing any existing margin calculations.
*  Do this without changing how loss socialisation works.

Roughly how:

* The insurance pool will become "margin account" for a "network party" (think of the network party as a "bad bank").
* Network will regularly check whether a party is distressed (but doesn't need to be tied to orders / trades etc. only at desired frequency e.g. every block or every 5s). If a party is in breach of margin level their margin is transferred to insurance pool and their position to the "network party". (*)
* Network does mark-to-market as usual. If the network party has a position the insurance pool is used for MTM flows. If insurance pool is empty network loss socialises as before
* The network party (insurance pool party) will employ a simple heuristic according to which it will attempt, at a desired intensity, to reduce it's position.

Insurance party position reduction heuristic:

* if market in continuous trading, place market orders once every desired number of seconds with size the smaller of "half available book volume", "half the position to close out" (unless the position is exactly 1 in which case close out all.
*  if market is in an auction, place a limit order - more thinking needed about the size / price; e.g., don't want to trigger price monitoring or move price too much.

(*) An added benefit of this is that the network may avoid a bunch of close-out trades - imagine network party is long 10 but now it is assigned a position of a party that was short 8; its new position is 2 ... and a bunch of network trades didn't need to happen.

## Specs
- [Spec issue](https://github.com/vegaprotocol/specs/issues/1641)

candida-d:
Closes #812


docs/concepts/trading-on-vega/market-protections.md
```diff
@@ -112,27 +112,27 @@ If a market enters into a liquidity auction and never again attracts enough liqu
...
 
 ### Closeouts
-When a participant does not have enough collateral to hold their open positions, the protocol will automatically trigger a closeout.
+When a participant does not have enough collateral to hold their open positions, the protocol will automatically trigger a closeout. The closeout process is a last resort for a position. 
 
-The closeout process is a last resort for a position. If a trader's deployed margin on the market is insufficient to cover a mark to market settlement liability, first Vega will search the trader's available balance of the settlement asset. If this search is unable to cover the full liability, the trader will be considered distressed and undergo position resolution. Any margin balance remaining after closeout is transferred to the market's insurance pool.
+If a trader's deployed margin on the market is insufficient to cover their mark to market settlement liability, first Vega will search the trader's available balance of the settlement asset. If the trader's full liability still can't be covered, their position is transferred to the network. Any margin balance remaining after their closeout is transferred to the market's insurance pool.
```
witgaw, line 122: That search is only done in cross-margin mode.

witgaw (commented):
Looks good to me, just some minor comments.

docs/tutorials/proposals/new-market-proposal.md
```diff
@@ -168,6 +168,16 @@ The risk model uses the following properties:
...
 
+### Liquidation strategy
+Set up the liquidation strategy to minimise the impact of distressed traders on the market. These parameters can balance between minimising the market impact of disposing of distressed positions and not holding a large open volume for a long time.
+
+| Field | Description | Suggested value |
+| ----------- | ----------- | ----------- |
+| `disposalTimeStep` | Interval, in seconds, at which the network will attempt to close a position it's acquired from distressed traders. | 30 |
+| `disposalFraction` | Fraction of the open position the market will try to close in a single attempt. Range 0 through 1 | 0.1 |
+| `fullDisposalSize` | Size of the position that the network will try to close in a single attempt  | 0 |
+| `maxFractionConsumed` | Maximum fraction of the order book's total volume, within the liquidity bounds, that the network can use to close its position. Range 0 through 1 | 0.1 |
```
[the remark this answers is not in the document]
candida-d: Thanks! I just pulled these values from a random testnet market and forgot to check for sensible ones.

docs/concepts/trading-on-vega/market-protections.md
candida-d, line 122:
```suggestion
How a trader's position is treated in a situation where they have insufficient deployed margin to cover their mark to market settlement liability depends on the margin mode they're using. 

If a trader using isolated margin doesn't have enough deployed margin on the market, then the position is closed out. Any margin balance remaining after their closeout is transferred to the market's insurance pool.

If a trader using cross margining doesn't have enough deployed margin on the market, first Vega will search the trader's available balance of the settlement asset. If the trader's full liability still can't be covered, their position is transferred to the network. Any margin balance remaining after their closeout is transferred to the market's insurance pool.
```

candida-d: @witgaw Could you sense check my edit here? The one thing I couldn't confidently confirm is what happens to the bits of margin leftover after a closeout in isolated margin.

docs/tutorials/proposals/new-market-proposal.md
```diff
@@ -168,6 +168,16 @@ The risk model uses the following properties:
...
 | `param: sigma` | Annualised historic volatility of the underlying asset. <br/><br/>Accepted values: any strictly non-negative real number; suggested value: asset dependent, should be derived from the historical time-series of prices. | 0.8 (converts to 80%) |
 
+### Liquidation strategy
+Set up the liquidation strategy to minimise the impact of distressed traders on the market. These parameters can balance between minimising the market impact of disposing of distressed positions and not holding a large open volume for a long time.
+
+| Field | Description | Suggested value |
+| ----------- | ----------- | ----------- |
+| `disposalTimeStep` | Interval, in seconds, at which the network will attempt to close a position it's acquired from distressed traders. | 30 |
+| `disposalFraction` | Fraction of the open position the market will try to close in a single attempt. Range 0 through 1 | 0.1 |
+| `fullDisposalSize` | Size of the position that the network will try to close in a single attempt  | 0 |
```
candida-d, line 178:
```suggestion
| `fullDisposalSize` | Size of the position that the network will try to close in a single attempt  | 1 |
```

docs/tutorials/proposals/new-market-proposal.md
candida-d, line 179:
```suggestion
| `maxFractionConsumed` | Maximum fraction of the order book's total volume, within the liquidity bounds, that the network can use to close its position. Range 0 through 1 | 0.05 |
```

witgaw (approved):
LGTM

PR #925, opened 2024-02-13
feat: add 0.74 release summary notes

gordsport:
Adds 0.74 release summary notes

Dates added are STC


gordsport:
@candida-d 

I think that the 0.74 brach this PR gors into may have gone. If there is a better place to push this to I will update

candida-d (approved):

PR #943, opened 2024-02-22
remove link to rewards page on website

rebeccajsimmonds:
this page does not exist any more


candida-d:
Replacing with a different PR because of unrelated build issues

PR #962, opened 2024-02-27
fix/0.74 proposal fields

edd:
- **add fundingRate fields**


candida-d:
@edd could you add these to the versioned docs too?

edd:
> @edd could you add these to the versioned docs too?
They were added in https://github.com/vegaprotocol/documentation/pull/962/commits/9340837df0677b8a3d11fc4f660f3bb39ab54524, so they *should* be there. Maybe there was a slow build?

candida-d (approved):

PR #996, opened 2024-03-15
Upgrade info for 0.75

candida-d:


docs/node-operators/migration-guides/upgrade-node.md
```diff
@@ -48,138 +48,13 @@ Add the following new parameters to the configuration files.
 
 We recommend checking all of the changes on your own. Follow the below instructions to do it, and read through every description to understand the changes:
 
-1. Download vega `v0.74.1`
+1. Download vega `v0.75.0`
 2. Generate config files within the temp home directory
   - vega config
   - comet BFT config
   - data-node config
 3. Compare the new generated file in the temp location and the old file to see the differences.
```
daniel1302, line 56: There is one parameter added(`NodeVoteResendInterval`). But We recommend it to 10s(the value won't change behavior - it was already hardcoded to 10s), but I think it's fine because.

daniel1302 (approved):

PR #1024, opened 2024-04-09
Spot markets

candida-d:


docs/concepts/liquidity/provision.md
```diff
@@ -82,6 +82,21 @@ The bond total will remain in the bond account for as long as the LP chooses to
...
+### Liquidity commitment for a spot market
+In almost all ways, providing liquidity to a spot market is the same as for other markets. There are minor differences when considering the commitment and how much to have available to support a spot market commitment. 
+
+Providing liquidity to a spot market requires setting aside a set amount of commitment as bond, using the market's quote asset. The orders for the commitment need to provide the full amount of assets required to support the liquidity for both the buy and sell side.
+
+#### Example
+
+An LP commits 100,000 USDT to the bond account for a BTC/USDT market.
+
+That LP has committed to providing 100,000 USDT worth of buy orders and sell orders.
```
cdummett, line 94: Small nitpick as I think worth avoiding the assumption that a commitment amount of 100,000 USDT results in requiring 100,000 USDT worth of orders. This won't always be the case due to the bond scaling.

Maybe just...
```suggestion
An LP has submitted a commitment and has an obligation to provide 100,000 USDT worth of buy orders and sell orders.
```
Want to avoid

docs/concepts/trading-on-vega/market-types.md
```diff
@@ -64,4 +64,11 @@ The margin calculations for perpetual futures markets contain an additional para
 * [Concept: Periodic settlement](./settlement.md#periodic-settlement-for-perpetuals)
 * [Concept: Positions and margin](./margin.md)
 * [Tutorial: Propose a perpetuals market](../../tutorials/proposals/new-perpetuals-market.md)
-:::
\ No newline at end of file
+:::
+
+## Spot
+Spot markets allow for buying and selling assets and receiving them as soon as the order is filled - in effect swapping one asset for another. You'll need the full amount of the asset you want to exchange, and will receive the full exchange amount in return. Spot markets are not margined, which means you also can't hold a position because as soon as your order is filled, it's completed.
```
cdummett, line 70: Maybe worth clarifying a user still pays fees on a spot market?

This line...

> and will receive the full exchange amount in return

to me suggests there are no fees.

candida-d, line 70:
```suggestion
Spot markets allow for buying and selling assets and receiving them as soon as the order is filled - in effect swapping one asset for another. You'll need the full amount of the asset you want to exchange, and will receive the full exchange amount, minus fees, in return. Spot markets are not margined, which means you also can't hold a position because as soon as your order is filled, it's completed.
```

docs/concepts/assets/accounts.md
```diff
@@ -62,6 +62,11 @@ Note that this happens instantaneously, so you will likely never observe a curre
 [Concept: Mark-to-market settlement](../trading-on-vega/settlement.md#mark-to-market-settlement) 
 :::
 
+### Holding accounts
+Holding accounts, like settlement accounts, temporarily holds assets to be distributed based on the outcomes of trades and market movements. However, holding accounts are specific to spot markets.
+
+An order on a spot market that doesn't immediately trade in full, has the collateral needed to cover the order, and its fees, moved to a holding account, to ensure you can support your open order. If you cancel or reduce the size, any freed up collateral is returned to your general account. This is true in both continuous trading and auction modes.
```
cdummett, line 68:
```suggestion
An order on a spot market that doesn't immediately trade in full, has the collateral needed to cover the order, and any possible fees, moved to a holding account, to ensure you can support your open order. If you cancel or reduce the size, any freed up collateral is returned to your general account. This is true in both continuous trading and auction modes.
```

docs/concepts/assets/accounts.md
```diff
@@ -62,6 +62,11 @@ Note that this happens instantaneously, so you will likely never observe a curre
 [Concept: Mark-to-market settlement](../trading-on-vega/settlement.md#mark-to-market-settlement) 
 :::
 
+### Holding accounts
+Holding accounts, like settlement accounts, temporarily holds assets to be distributed based on the outcomes of trades and market movements. However, holding accounts are specific to spot markets.
```
cdummett, line 66: Not sure how I feel about this line, don't think it adds much and possibly just makes a confusing link to settlement accounts?

cdummett (approved):

PR #1050, opened 2024-05-09
Update testnet release notes for 0.76.1

edd:
Merges pre-release 0.76.0.preview-1 and -preview-3 in to a full 0.76.1 set of release notes


edd:
Closing in favour of pushing these release notes in to #1049 and getting both updated at the same time

PR #1052, opened 2024-05-14
feat: update from and to blocks range in the arbitrum rpc query to hex ones

daniel1302:
Some of the RPC endpoints return  the following error when values for `fromBlock` and `toBlock` are not prefixed with 0x:

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32602,"message":"invalid argument 0: hex string without 0x prefix"}}
```


candida-d (commented):
@daniel1302 it looks good! But could you also make the same change on the “get started” page in mainnet?

candida-d (approved):

PR #1089, opened 2024-07-12
Guide for upgrading to 0.77

candida-d:


daniel1302 (approved):
Looks good, I have compared new parameters and there are no parameters We should mention in the docs, there were new params in tm and vega, but people can use default version

PR #1092, opened 2024-07-16
Update websocket.md

powrusr:
That extra 's' caused me some troubleshooting pain yesterday. I copied the endpoints from here. Thanks for the websocat pointer Barnabee.


candida-d:
Thanks for contributing to fixing this page, @powrusr! 

On your branch, could you also apply this change to the file in this directory? https://github.com/powrusr/vegadocs/blob/patch-1/versioned_docs/version-v0.76/api/websocket.md That will update it for the 'mainnet' version of the documentation.

powrusr:
Okay. I'll try to remember to do that in the future for both testnet and mainnet at once. Greets.

candida-d (approved):

PR #1114, opened 2024-08-19
Rename reward + new type

candida-d:


docs/concepts/trading-on-vega/discounts-rewards.md
```diff
@@ -65,6 +65,14 @@ All of the details for the volume discount program are proposed and accepted thr
 * [Spec: Technical design of the volume discount program ↗](https://github.com/vegaprotocol/specs/blob/master/protocol/0084-VDPR-volume_discount_program.md).
 :::
 
+### Rewards for eligible participants
+Participants who exceed certain eligibility criteria can be rewarded when the eligible entities reward is funded. 
+
+When setting up the reward, one or both criteria can be set:
+
+* Minimum number of VEGA tokens that need to be staked
+* Minimum notional time-weighted average position, on any market using the asset nominated in the reward, or the specified market/markets
```
cdummett, line 74: We technically also would allow neither to be set, if a markets scope is specified.

It's use-case would be reward anyone trading on one of these markets, niche and game-able so maybe not worth mentioning.

docs/concepts/trading-on-vega/discounts-rewards.md
```diff
@@ -159,9 +167,11 @@ Since rewards can only be provided if they're funded, the [recurring transfer](.
 
 **Pro-rata**: A participant's reward is scaled based on their score.
 
-**Rank**: A participant's reward is scaled based on where their score lands on the rank table. The rank table determines a rank group, for example 1-8, and what ratio of the reward amount that group would receive.
+**Rank**: A participant's reward is scaled based on where their score lands on a rank table, which is also set in the transfer. The rank table determines a rank group, for example 1-8, and what ratio of the reward amount that group would receive.
+
+**Lottery**: Each eligible participant is randomly assigned a position in the rank table. If the lottery distribution is used for a trading reward (i.e., s not the eligible entities reward), the probability of a participant being selected is based on their score towards that trading reward.
```
cdummett, line 172: Not sure that `s` is supposed to be there?
```suggestion
**Lottery**: Each eligible participant is randomly assigned a position in the rank table. If the lottery distribution is used for a trading reward (i.e., not the eligible entities reward), the probability of a participant being selected is based on their score towards that trading reward. 
```

docs/concepts/trading-on-vega/discounts-rewards.md
```diff
@@ -212,10 +222,10 @@ The total maker fees paid by all parties in that market is *$10,000*.
...
 
-### Largest traders by position size
-The largest traders by position size category rewards traders with consistenly larger positions that rank higher in the standings than other traders, as long as they can keep the positions open.
+### Largest traders by notional size
+The _largest traders by notional size_ category rewards traders whose positions have consistently higher notional that rank higher in the standings than other traders, as long as they can keep the positions open.
 
-It measures a trader's time-weighted average position over a set number of epochs to determine how long each trader is able to manage a position that's larger than the positions of other traders, without being closed out. It's also known as the "average position metric".
+It measures a trader's time-weighted average position over a set number of epochs to determine how long each trader is able to manage a position that's larger than the positions of other traders, without being closed out. Your final average position size will be multiplied by the mark price at the end of the epoch to allow positions across different markets to be compared. It's also known as the "average notional metric".
```
cdummett, line 228: This looks like something I have said at some point but is no longer true - where did this come from?

cdummett:
```suggestion
It measures a trader's time-weighted average notional position over a set number of epochs to determine how long each trader is able to manage a position that's larger than the positions of other traders, without being closed out. It's also known as the "average notional metric".
```

PR #1136, opened 2024-09-24
new wallet software

candida-d:


docs/tools/index.md
```diff
@@ -14,5 +14,11 @@ There are several ways to interact with a Vega Wallet:
 ## Trading interface
 Use **[Vega Console](https://console.fairground.wtf)** to try trading on a testnet trading interface. The Fairground network is set up to provide a fairly realistic experience where you can try out the Vega software to place orders, provide liquidity, and see how the market mechanics work.
 
+## Governance interface
+Vega has open-source software for creating and customising a governance user interface. You can find the code on [GitHub](https://github.com/vegaprotocol/frontend-monorepo/tree/develop/apps/governance).
+
+## Block explorer interface
+Vega has open-source software for creating and customising a block explorer user interface. You can find the code on [GitHub](https://github.com/vegaprotocol/frontend-monorepo/tree/develop/apps/explorer).
```
edd, line 21: **thought (non-blocking)**: Governance uses the standard data nodes, while Explorer relies on an extra Block Explorer service to be run (https://github.com/vegaprotocol/vega/tree/develop/blockexplorer, and it is probably elsewhere in the docs). It's possibly not totally relevant, but sprung to mind because this and the governance interface paragraphs are so similar, that some extra detail makes it feel less copy & pastey

edd (approved):
One not-very-important suggestion, otherwise fine

PR #1147, opened 2024-11-06
Vaults

candida-d:


docs/concepts/assets/accounts.md
```diff
@@ -24,6 +24,35 @@ Any assets that are in a general account can be withdrawn or transferred.
 **[REST](../../api/rest/data-v2/trading-data-service-get-party.api.mdx)**: Use the API to check your public key for the accounts your assets are allocated to.
 :::
 
+## Vault
+A *vault* holds assets that have been contributed by one or more parties to be used by another party for trading. The assets in a vault can only be used to enter and exit positions. Vaults have general and margin accounts associated with them.
```
ze97286, line 28: I would say maybe by the vault owner rather than by some other party?

docs/concepts/assets/accounts.md
```diff
@@ -24,6 +24,35 @@ Any assets that are in a general account can be withdrawn or transferred.
...
+
+Contributors to a vault have a share of the value in the vault, based on how much their deposited amount contributed to its original value.
+
+The vault account manager - the party that created the account - can charge fees on account withdrawals if they choose to.
```
ze97286, line 32: is account manager how we want to call the party who created the vault?

ze97286: > can charge fees on account withdrawals if they choose to.
not on withdrawals, fees are on performance and on management not on withdrawals.

ze97286: you may also want to add that the vault owner (account manager) can transfer ownership of the vault.

docs/concepts/assets/accounts.md
```diff
@@ -24,6 +24,35 @@ Any assets that are in a general account can be withdrawn or transferred.
...
+
+Each vault account handles one asset.
+
+### Contribute to a vault
+To deposit assets into a vault, a participant will need to submit a `depositIntoVault` transaction which includes the vault ID (account ID), and the amount to contribute.
+
+### Redeem from a vault
```
ze97286, line 39: I think you want to add information on how to actually withdraw. This below explains how the mechanics of redemption works on the vault side, but not that I need to send a transaction saying how much I want to withdraw etc.

ze97286: also worth mentioning the cutoff that is added to the withdrawal time.

docs/concepts/assets/accounts.md
```diff
@@ -24,6 +24,35 @@ Any assets that are in a general account can be withdrawn or transferred.
...
+Each vault has a list of dates when redemption takes place.
+
+There are two redemption types:
+- Available funds only: Vault participants can redeem from assets available in the vault's general account. When a withdrawal request falls closest to an `available funds only` date, a fraction can be withdrawn. The withdrawal amount will be the lesser of the amount requested, and the amount that available to withdraw based on the participant's share of the 'maximum redemption fraction * available amount'. Each redemption is fulfilled as much as possible, and the redemption request is considered complete. Any remainder of the redemption requests is cancelled and participants can request a redemption for the next date.
+- Normal: Vault participants can redeem up to a certain fraction of the vault's total balance, which includes the general and margin accounts. If any amount of the withdrawal cannot be fulfilled, the remaining is marked as 'late', and whenever more assets reach the general account, late withdrawals will be fulfilled.
```
ze97286, line 44: I would say "and whenever more funds" rather than "and whenever more assets"

docs/concepts/assets/accounts.md
```diff
@@ -24,6 +24,39 @@ Any assets that are in a general account can be withdrawn or transferred.
...
+
+There are two redemption types:
+- Available funds only: Vault participants can redeem from assets available in the vault's general account. When a withdrawal request falls closest to an `available funds only` date, a fraction can be withdrawn. The withdrawal amount will be the lesser of the amount requested, and the amount that available to withdraw based on the participant's share of the 'maximum redemption fraction * available amount'. Each redemption is fulfilled as much as possible, and the redemption request is considered complete. Any remainder of the redemption requests is cancelled and participants can request a redemption for the next date.
+- Normal: Vault participants can redeem up to a certain fraction of the vault's total balance, which includes the general and margin accounts. If any amount of the withdrawal cannot be fulfilled, the remaining is marked as 'late', and whenever more funds reach the general account, late withdrawals will be fulfilled.
+
+Each redemption date also has a cut-off period beforehand, set by the vault manager. Any redemption requests that come during the cutoff period are queued until the following redemption date.
```
ze97286, line 48: not sure it needs to be said here but the cutoff is defined for the vault, not different for each redemption date.

docs/concepts/assets/accounts.md
```diff
@@ -24,6 +24,39 @@ Any assets that are in a general account can be withdrawn or transferred.
 **[REST](../../api/rest/data-v2/trading-data-service-get-party.api.mdx)**: Use the API to check your public key for the accounts your assets are allocated to.
 :::
 
+## Vault
```
ze97286, line 27: what about documenting the vault management operations, create/update/change ownership?

docs/concepts/assets/accounts.md
```diff
@@ -24,6 +24,53 @@ Any assets that are in a general account can be withdrawn or transferred.
...
+The transaction needs to include the following information for the vault:
+- Settlement asset for the vault.
+- Metadata describing the vault.
+- Fee period, which sets the frequency of the vault's fees assessment
+- Management fee factor.
+- Performance fee factor.
+- Redemption dates
+- The cutoff period following the redemption date until which a redemption request is queued
+
+Once a vault is created, if the vault manager wants to transfer ownership to another party, they'll need to submit a `ChangeVaultOwnership` transaction with the vault ID and the Vega public key of the new vault manager.
```
ze97286, line 48: you may want to include something about update vault here too, including the rules for what can and cannot change - especially on redemption dates.

